PR #16, opened 2023-03-20
enable format flags for cluster list

kyzrfranz:
Just like microceph, to round it up


microovn/cmd/microovn/cluster_list.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"github.com/lxc/lxd/shared/i18n"
```
stgraber, line 5: This import should be moved to the second chunk of imports.

We usually use this grouping:
```
built-in

external imports

internal imports (github.com/canonical/microovn/XYZ in this case)
```

kyzrfranz: yeah sry, that was my IDE i guess

PR #26, opened 2023-05-12
Ensure that joining member finishes its  `ovn.Join` execution before rest of the cluster is notified.

mkalcok:
New feature in `microcluster` splits `OnJoin` hook into `PreJoin` and `PostJoin`. `PreJoin` hook is guaranteed to execute before rest of the cluster members execute their `OnNewMember` hooks.

This fixes a bug in which existing cluster members execute `ovn.Refresh` before the new member registers its services to the dqlite database.


fnordahl (commented):
nit: the length of the commit subject and the individual lines in the message is a bit excessive.

I tend to try to limit the length so that it looks nice in `git log` in a 80 column terminal.

PR #29, opened 2023-05-30
snap: Stabilization patch for 22.03

fnordahl:
Lock down parts built from git.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -125,6 +125,7 @@ parts:
       - raft
     source: https://github.com/canonical/dqlite
     source-type: git
+    source-branch: v1.15.1
     source-depth: 1
     plugin: autotools
     autotools-configure-parameters:
@@ -145,6 +146,7 @@ parts:
   raft:
     source: https://github.com/canonical/raft
     source-type: git
+    source-branch: v0.17.1
     source-depth: 1
     plugin: autotools
     autotools-configure-parameters:


snapcraft.yaml
```diff
@@ -159,6 +161,19 @@ parts:
...
+            dpkg-deb -f \
+            $CRAFT_PART_SRC/../stage_packages/ovn-host*.deb \
+            Version | sed -rne 's/([0-9.]+)[-+].*$$/\1/p')
+        git_version=$(
+            git -C $CRAFT_PROJECT_DIR describe \
+                --always \
+                --dirty \
+                --tags \
+                --abbrev=10 | sed -s 's/^v//')
+        craftctl set version=${pkg_version}+${git_version}
```
fnordahl, line 176: As @freyes pointed out, using the Git sha from a different project than the first part of the version might be confusing and not what the idle observer might expect.

Perhaps we need to prefix it with something, i.e. `22.03.0+microovn6001c4d922`, it becomes a bit long though.

We need to bear in mind that the maximum length of the version string is 32 characters ([relevant discussion](https://forum.snapcraft.io/t/feature-request-extend-length-of-version-string/13335))

Thoughts @dshcherb @mkalcok @sabaini ?

sabaini: Yep, with `microceph` in there it'd be even longer. 

Some options that come to mind

1. Document the semantics of the version string somewhere else so at least ppl can look it up
2. Use some mnemonic, e.g. `22.03.0+mo6001c4d922` for `microovn` 
3. Or alternatively a mnemonic making it clearer the snap git has is referenced `22.03.0+snap6001c4d922`

Nothing's ideal but IMHO 3. might be the least worst of the ones above?

fnordahl: Yes, I like alternative 3 as well, it's both shorter and applies equally well to both (and other similar) projects.

mkalcok: I'm just thinking out loud here, but do we need to involve version from ovn package at all? We could just add tag to this repository and use
```
git describe --tags --always HEAD
```
 to generate the version.
 Until we add first tag, this command will output just hash which is imo fine, it's just temporary. Then lets say we tag our first release with `22.03`, the command above will output just `22.03` which looks good in the snap store. And then when we push commits after the tag we'll get `<tagname>-<ncommits-ahead>-<commit-id>`

fnordahl: Thank you for providing your thoughts @mkalcok!

Assuming we would want to tag the microovn repository following the upstream versions, your suggestion is not unreasonable, and my question then becomes; is that what we want to do?

Snapcraft provides us with the means to declare a set of parts to put into a package. We can either source parts from the Ubuntu distribution and associated archives of debs (such as UCA), or source parts directly from upstream Git repositories.

In both cases we will be able to consume a versioned artifact that has already received release engineering from either of those upstream projects.

The microovn snap is in essence a deployment tool, simpler and still similar to charms. From previous experience with the OVN charms, there is not much change in the deployment specific parts for each upstream release.

My reasoning behind building the version based on the upstream git/package version + some other version for the snap part is that I expect the upstream version to change much more often than the corresponding snap version. If we were to tag the snap project for every upstream version, I think we will be inventing unnecessary busywork for ourselves.

The need for exposing the snap version comes from changes we make to the snap specific code, which may change within the same upstream OVN version, and we may want to know which version of the snap code a user is running whenever an issue is reported.

freyes: > The microovn snap is in essence a deployment tool, simpler and still similar to charms

although from a QA perspective you would want to lockdown the snap to a specific ovn version, if the deb package moves to a new ovn version you would want to (re)validate the snap when you are rebuilding with this new version, and at that point do the tagging sounds reasonable.

Depending on how your CI/CD pipeline looks like you could do the tagging at that point automagically.

I think in the end we would want to make the conscious decision if the snap is loosely coupled to the ovn version (AKA it has code to adapt itself to the version of ovn it's carrying) or not, for example in the sunbeam world the charm and the rock are strongly coupled

fnordahl: > > The microovn snap is in essence a deployment tool, simpler and still similar to charms
> 
> although from a QA perspective you would want to lockdown the snap to a specific ovn version, if the deb package moves to a new ovn version you would want to (re)validate the snap when you are rebuilding with this new version, and at that point do the tagging sounds reasonable.

We definitively want validation for any new major version, and I think that a new major version will be expressed as a new track in the snap store.

Our current pipeline is based on [Launchpad snap recipes](https://launchpad.net/microovn/+snaps), which consumes the code from branches or tags to build a certain track.

Now, if we for example feed a branch into the Launchpad snap recipe machinery, locking the OVN part down to whatever is available in the base Ubuntu distribution defined for that branch, we would be a beneficiary of all the work that goes into that Ubuntu distribution throughout its lifetime, including security and point release updates. Ideally, those kinds of updates can just flow through without manual intervention, at least to the edge channel.

Would you think that additional tagging of the snap git repository be necessary in this context?

freyes:
On Wed, 2023-05-31 at 11:53 -0700, Frode Nordahl wrote:
> 
> Now, if we for example feed a branch into the Launchpad snap recipe machinery, locking the OVN
> part down to whatever is available in the base Ubuntu distribution defined for that branch, we
> would be a beneficiary of all the work that goes into that Ubuntu distribution throughout its
> lifetime, including security and point release updates. Ideally, those kinds of updates can just
> flow through without manual intervention, at least to the edge channel.
> Would you think that additional tagging of the snap git repository be necessary in this context?

In my ideal world, yes, a new tag, but not made by a human, by the bot that did the rebuild. I think
the string "23.03.1-ubuntu1+snap02a345" suggested by Peter is a good compromise.

fnordahl:
> On Wed, 2023-05-31 at 11:53 -0700, Frode Nordahl wrote: Now, if we for example feed a branch into the Launchpad snap recipe machinery, locking the OVN part down to whatever is available in the base Ubuntu distribution defined for that branch, we would be a beneficiary of all the work that goes into that Ubuntu distribution throughout its lifetime, including security and point release updates. Ideally, those kinds of updates can just flow through without manual intervention, at least to the edge channel. Would you think that additional tagging of the snap git repository be necessary in this context?
> In my ideal world, yes, a new tag, but not made by a human, by the bot that did the rebuild. I think the string "23.03.1-ubuntu1+snap02a345" suggested by Peter is a good compromise.

Automatic tagging from a system would indeed be nice. The snap consumes packages from Ubuntu, and we have an automatic build pipeline set up with Launchpad. The version would differ depending on what packages are available in the source Ubuntu distribution, and may change without the snap repository changing. So I guess we would need to defer availability of such automatic tagging from Launchpad at some point in the future.

As for the version string, I'd rather stick to the upstream version part of the package version. The reason for this is that it is technically possible to install this snap on a non-Ubuntu system, and having the `ubuntu` string as part of the version would be confusing for said users.

Example:
```
[root@f38 ~]# lsb_release -a
LSB Version:	:core-4.1-amd64:core-4.1-noarch
Distributor ID:	Fedora
Description:	Fedora release 38 (Thirty Eight)
Release:	38
Codename:	ThirtyEight
[root@f38 ~]# snap install microovn
2023-08-07T14:05:24Z INFO snap "microovn" has bad plugs or slots: microovn (unknown interface "microovn")
microovn 0+git.f8a4497 from Canonical✓ installed
WARNING: There is 1 new warning. See 'snap warnings'.
[root@f38 ~]# microovn cluster bootstrap
[root@f38 ~]# microovn status
MicroOVN deployment summary:
- f38 (10.215.82.189)
  Services: central, chassis, switch
```

PR #30, opened 2023-05-30
Switch to in-db configuration of OVN remotes.

mkalcok:
By configuring "remotes" in database, rather than in CLI arguments, we can change values during runtime without restarting the service.

This proposal is part of a bigger change to switch to using tls for OVN communication instead of plaintext, but I thought that it's better to split it into multiple smaller PRs.

Changes included:
* Setting `--db-{nb,sb}-create-insecure-remote=no` in our startup scripts, prevents `ovn-ctl` to hardcode "--remote" configuration on CLI and we can configure the values in database without the need to restart service on change.
* Added in-code calls to `ovn-{nb,sb}ctl` to configure remotes in database*
* New module `commands.go` that contains wrappers for execution of shell commands like `ovs-vsctl`, `ovn-nbctl` ...
  * Reasoning for wrappers: We execute these commands fairly soon after starting/restarting of their respective services and I encountered occasional race condition when the commands got executed before the service was ready. The wrappers utilize `ovsdb-client wait` to ensure that the target database is ready before we execute the main command. (nice feature of the "wait" command is that it wont fail even if the targeted socket does not exist yet, it will wait for its eventual creation)
  * Another side-effect is that we also got rid of some (not all) calls to [deprecated `RunCommand()` from lxd package](https://github.com/lxc/lxd/blob/ef84f7b00743aa1cbff462e8b6044487229fadd0/shared/util.go#L987)
  
  *Note: The in-db configuration of "remotes" is shared between all cluster member. Adding listening configuration with specific IPs results in every member excessively complaining about inability to bind to the IP that does does not belong to the node. As a solution we decided to set configuration that listens on all IPs (e.g. `ptcp:6641`). This is bit of a step back while we figure out how to make [`Local_Config` database](https://github.com/openvswitch/ovs/commit/6f24c2bc769afde0a390ce344de1a7d9c592e5a6#diff-b348bd750b5e924aca31a8629de792541158a812a5c16855fe2dbdd688ad25ee) work in the MicroOVN setup. In the meantime, access control can still be enforced externally, for example by using firewall.


microovn/ovn/commands.go
```diff
@@ -0,0 +1,150 @@
...
+
+// newOvsdbSpec is a helper function that takes OvsdbType as an argument and generates
+// proper ovsdbSpec for given type.
+func newOvsdbSpec(dbType OvsdbType) (*ovsdbSpec, error) {
+	var dbSpec *ovsdbSpec
+	var err error
+
+	if dbType == OvsdbTypeNB {
+		dbSpec = &ovsdbSpec{
+			Path: filepath.Join(os.Getenv("SNAP_COMMON"), "run", "central", "ovnnb_db.sock"),
```
fnordahl, line 41: The snap controls where OVS/OVN payload puts its files and sockets through environment variables, for example:
https://github.com/canonical/microovn/blob/a25b61d001db11d36800cbf1d39a4c4c7d0e493f/snapcraft/commands/central.start#L8-L12

We can't use that environment variable in the Go code directly, as the Go code manages multiple executables, each with their own environment. At the same time I don't feel comfortable also hard coding the same path in the Go code individually, would there be a way to provide these paths from a single source of truth?

At least we should start with a constant for the strings being repeated in the code?

snapcraft/commands/chassis.start
```diff
@@ -22,10 +22,12 @@ done
 export PATH="${OVN_RUNDIR}/bin/:${PATH}"
 
 # Prepare the arguments
+# By specifying "--db-*-create-insecure-remote=no" we prevent creation of
+# hardcoded bindings and we can use database to configure remotes later.
 OVN_ARGS="--db-nb-addr="${OVN_LOCAL_IP}" \
---db-nb-create-insecure-remote=yes \
```
fnordahl, line 26: Not a problem of this patch, but why do we even list these arguments when starting the controller?

AFAICT none of these are required as the `ovn-controller` will get its information from the local Open vSwitch DB

microovn/ovn/commands.go
```diff
@@ -0,0 +1,150 @@
...
+	"strconv"
+)
+
+const defaultDBConnectWait = 30 //Default time to wait for connection to ovsdb
+
+// ovsdbSpec is a helper structure for precise identification of ovsdb databases. A lot of
+// ovn/ovs commands take path to either database file, database socket or process control socket
+// along with the database name. This structure can be used for such cases.
+type ovsdbSpec struct {
+	Path string
```
fnordahl, line 19: The use of the keyword `Path` does not resonate with me. The tools you are modelling can accept many types of connection methods which may be represented as paths, remote IP addresses or something else.

PS: This is just a command wrapper so we do not yet have a lot of requirements for it, but just wanted to make you aware of the existence of https://github.com/ovn-org/libovsdb

microovn/ovn/bootstrap.go
```diff
@@ -77,14 +74,47 @@ func Bootstrap(s *state.State) error {
...
+	nbDB, err := GetOvsdbLocalPath(OvsdbTypeNB)
+	if err != nil {
+		return fmt.Errorf("Failed to get path to OVN NB database socket: %w", err)
+	}
+	sbDB, err := GetOvsdbLocalPath(OvsdbTypeSB)
+	if err != nil {
+		return fmt.Errorf("Failed to get path to OVN SB database socket: %w", err)
+	}
+
+	_, err = NBCtl(
```
fnordahl, line 86: When connecting to the local DB socket, this command will only succeed when talking to the DB leader.

I see no check for whether this is the leader or not, would it be needed here?

microovn/ovn/bootstrap.go
```diff
@@ -77,14 +74,47 @@ func Bootstrap(s *state.State) error {
...
+		s,
+		fmt.Sprintf("--db=unix:%s", nbDB),
+		"set-connection",
+		"ptcp:6641",
+	)
+	if err != nil {
+		return fmt.Errorf("Error setting ovn NB connection string: %s", err)
+	}
+
+	_, err = SBCtl(
```
fnordahl, line 96: Same as above.

fnordahl (changes requested):
Thank you for your work on this, Martin!

I added some comments in-line.

microovn/ovn/bootstrap.go
```diff
@@ -4,10 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-
```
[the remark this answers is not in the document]
mkalcok: yup, i used default settings for `goimports`, didn't realize that there's an option to separate stdlib packages.

microovn/ovn/commands.go
mkalcok: Agreed. They're are already sort of defined in here
https://github.com/canonical/microovn/blob/a25b61d001db11d36800cbf1d39a4c4c7d0e493f/microovn/ovn/paths.go#L9-L23
they are just not exporter.

snapcraft/commands/chassis.start
mkalcok: I think you are right. I'll take them out and test that everything works as expected.

microovn/ovn/commands.go
mkalcok: > The use of the keyword Path does not resonate with me.

In the current context, the only "end-user" of this value was `ovsd-client wait` command intended to run against local target but you are right, the semantic should be improved. Perhaps we can rename it to `Target` and also rename the `OvsdbType` enum members from
```
OvsdbTypeNB
OvsdbTypeSB
```
to
```
OvsdbTypeNBlocal
OvsdbTypeSBlocal
```
to make it more explicit. And later when the need arises we can add new enum members with `remote` suffix and add logic for generating those value.

> PS: This is just a command wrapper so we do not yet have a lot of requirements for it, but just wanted to make you aware of the existence of https://github.com/ovn-org/libovsdb

I did not. thank you. This is good stuff, I'll look into whether it makes sense to start using it right now to remove some of the clunkyness. I'm also not a fan of invoking shell when it's not necessary.

microovn/ovn/bootstrap.go
mkalcok: If your comment is meant in the context of this specific call to `NBCtl` then I don't think any check is needed as this is a bootstrap operation where it's guaranteed that there are no other cluster members.

If, on the other hand, you are talking about the overall `NBCtl` design, the command does not make any assumptions about remote targets. User is responsible for passing all required arguments, including `--db`, correctly. If there's a problem with execution (e.g. due to executing against non-leader), it'll bubble up as a generic command error.

fnordahl (approved):
LGTM, thanks!

PR #31, opened 2023-06-02
Fix some race conditions

stgraber:
This avoids writing an invalid environment and also makes it so that restarting microovn will re-generate the environment file.


microovn/ovn/bootstrap.go
```diff
@@ -89,7 +89,7 @@ func Bootstrap(s *state.State) error {
 		s,
 		fmt.Sprintf("--db=unix:%s", nbDB),
 		"set-connection",
-		"ptcp:6641",
+		"ptcp:6641,[::]",
```
fnordahl, line 92: Have not checked, but would we not need to also specify connection type and port for the v6 listener? Something like `ptcp:6641:[::]`

tomponline: Yeah I think it should be colon not comma between port and address.

fnordahl: So the comma is to list multiple connections, the format for each connection is defined here: https://manpages.ubuntu.com/manpages/kinetic/man7/ovsdb.7.html

I'll check if `set-connection` actually accepts multiple, or if we need to add them in a different way

tomponline: Yeah we just dont need the comma. So `ptcp:6641:[::]"` rather than `ptcp:6641,[::]"

stgraber: Updated

tomponline (approved):

microovn/ovn/start.go
```diff
@@ -0,0 +1,45 @@
...
+		return err
+	}
+
+	// Re-generate the configuration.
+	err = generateEnvironment(s)
+	if err != nil {
+		return fmt.Errorf("Failed to generate the daemon configuration: %w", err)
+	}
+
+	// Reconfigure OVS to use OVN.
```
mkalcok, line 28: Is there a reason why we don't reconfigure OVN NB and SB as well here? Calling `NBCtl` and `SBCtl` here to configure listening address could help us smooth out the upgrade process going from "cli-based" to "in-db" listen configuration.

stgraber: I don't really like the idea of running commands on every start which we know should never ever change, that just ends up needlessly burning CPU and time on every startup for every user.

What we've done for situations like these in LXD is keep a list of `patches` that have been applied to a particular system, either in the database or in a text file on disk. On startup, we load up the list of patches that have been applied and compare it with the list of patches in the code, anything that's not been applied yet gets applied.

Those patches are simple Go functions and are in charge of dealing with things like relocating data or reconfiguring things, basically actions which must be performed on every system but only be performed once.

mkalcok: Oh yeah, makes sense since right now the listen string is pretty much static. I was thinking ahead to when we'll use `Local_Config` database and listen strings with static IP, but we'll cross that bridge when we'll get to it. :+1:

I think that for the "upgrade problem" we can just utilize [refresh hooks](https://snapcraft.io/docs/supported-snap-hooks#heading--post-refresh) from snap.

PR #33, opened 2023-06-12
snap: Add ``ovn-trace`` command

fnordahl:
The ``ovn-trace`` command is useful for end users of the snap for debugging configuration issues and/or CMS/OVN bugs.


mkalcok (approved):
LGTM, I tested the deployment and the `ovn-trace` seems to be properly exposed to user.

mkalcok (approved):
Thanks for integrating this feature. LGTM

PR #34, opened 2023-06-12
snap: Build ovs and ovn from git for edge channel?

fnordahl:
There are some benefits in tracking git for the edge channel: 1) We get early feedback on any breaking upstream changes. 2) Checking whether problems encountered in stable versions
   are already fixed in newer upstream versions become easier.

Some drawbacks are:
1) Before cutting a new stable release, we most likely want to
   switch to consuming ovs/ovn from a supported Ubuntu package
   (security, known debug symbols, etc), and switching between
   them is more work.
2) Using different source during development and in stable might
   lead to unpleasant surprises in the event something we relied
   on unexpectedly is not available in the stable upstream
   version.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -140,13 +140,52 @@ parts:
       - lib/*/libuv.so*
 
   ovn:
-    plugin: nil
-    stage-packages:
-      - ovn-central
-      - ovn-host
-    organize:
-      usr/bin/: bin/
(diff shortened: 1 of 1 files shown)


fnordahl:
This proposal has been superseded by #49, which proposes to make use of the `devel` build-base instead, which is tracking the current development release.

PR #35, opened 2023-06-20
Cluster downscaling

mkalcok:
This change implements downscaling of a MicroOVN cluster and consists of two main parts.

Fix for a database schema. Currently `services` table contains foreign key referencing `internal_cluster_members`. However there's no `ON DELETE CASCADE` so attempts to remove microOVN cluster member fails with `Error: Delete "internal_cluster_members": FOREIGN KEY constraint failed`. This  change adds a database migration that fixes this issue.

Implementation for OVN cluster departure. `PreRemove` hook is added that attempts to
- Gracefully stop the OVN Controller, therefore removing related Chassis from OVN SB database
- Gracefully leave OVN NB cluster
- Gracefully leave OVN SB cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,59 @@ microovn.ovn-sbctl show
 microovn.ovn-appctl -t /var/snap/microovn/common/run/central/ovnnb_db.ctl cluster/status OVN_Northbound
 ```
 
+## Downscaling
+
+> **Warning**
+>
+> Be aware of effect of downscaling on availability and resiliency of your
+> cluster, especially when you are removing members that run OVN central
+> service (OVN SB, OVN NB, OVN Northd). OVN uses
+> [Raft consensus algorithm](https://raft.github.io) for cluster management,
+> which has fault tolerance of up to `(N-1)/2` members. It means that if you
+> scale back 3 node cluster to just 2 nodes, you lose any fault resiliency.
+
+Removing MicroOVN cluster member is as easy as running
+
+```shell
+microovn cluster remove <member_name>
+```
+
+MicroOVN will stop and disable any chassis components running on the member
+(`ovn-controller` and `ovs-vswitchd`), before attempting to gracefully leave
+any OVN database clusters for members with central components present. You can
+watch logs on departing member for any indications of removal failures with:
+
+```shell
+snap logs -f microovn.daemon
+```
+
+If there are any issues during the removal process, you will have to take
+manual steps to fix the OVN cluster.
+
+After the removal, you can also check the state of OVN services, to make sure
+that the member was properly removed.
+
+```shell
+# Check status of OVN SB cluster
+microovn.ovn-appctl -t /var/snap/microovn/common/run/central/ovnsb_db.ctl cluster/status OVN_Southbound
+
+# Check status of OVN NB cluster
+microovn.ovn-appctl -t /var/snap/microovn/common/run/central/ovnnb_db.ctl cluster/status OVN_Northbound
+
+# Check registered chassis
+microovn.ovn-sbctl show
+```
+
+### Data preservation
+
+MicroOVN will back up selected data directories into the timestamped location
+`/var/snap/microovn/common/backup_<timestamp>/`. These backups will include:
+
+* Logs
+* OVN database files
+* OVS database file
+* Issued certificates and keys
+
 ## TLS encryption
 
 MicroOVN enables SSL/TLS in OVN by default. It uses self-signed CA certificate
--- a/microovn/database/schema.go
+++ b/microovn/database/schema.go
@@ -12,6 +12,7 @@ import (
 // Each entry will increase the database schema version by one, and will be applied after internal schema updates.
 var SchemaExtensions = map[int]schema.Update{
 	1: schemaUpdate1,
+	2: schemaUpdateCascadeDeleteServices,
 }
 
 func schemaUpdate1(ctx context.Context, tx *sql.Tx) error {
@@ -32,6 +33,30 @@ CREATE TABLE services (
 );
   `
 
+	_, err := tx.Exec(stmt)
+
+	return err
+}
+
+// schemaUpdateCascadeDeleteServices ensures that records from 'services' are properly deleted
+// when associated 'internal_cluster_member' is removed.
+func schemaUpdateCascadeDeleteServices(_ context.Context, tx *sql.Tx) error {
+	stmt := `
+PRAGMA foreign_keys = OFF;
+CREATE TABLE services_new (
+  id                            INTEGER  PRIMARY KEY AUTOINCREMENT NOT NULL,
+  member_id                     INTEGER  NOT  NULL,
+  service                       TEXT     NOT  NULL,
+  FOREIGN KEY (member_id) REFERENCES "internal_cluster_members" (id) ON DELETE CASCADE
+  UNIQUE(member_id, service)
+);
+
+INSERT INTO services_new SELECT id, member_id, service FROM services;
+
+DROP TABLE services;
+ALTER TABLE services_new RENAME TO services;
+PRAGMA foreign_keys = ON;
+`
 	_, err := tx.Exec(stmt)
 
 	return err
--- a/microovn/ovn/certificates.go
+++ b/microovn/ovn/certificates.go
@@ -180,10 +180,13 @@ func DumpCA(s *state.State) error {
 	err = s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
 		CACertRecord, err = database.GetConfigItem(ctx, tx, CACertRecordName)
 		if err != nil {
-			return fmt.Errorf("failed to store CA certificate in the database: %s", err)
+			return fmt.Errorf("failed to get CA certificate from the database: %s", err)
 		}
 		return err
 	})
+	if err != nil {
+		return err
+	}
 
 	certPath := paths.PkiCaCertFile()
 	certFile, err := os.Create(certPath)
--- a/microovn/ovn/commands.go
+++ b/microovn/ovn/commands.go
@@ -12,6 +12,8 @@ import (
 )
 
 const defaultDBConnectWait = 30 //Default time to wait for connection to ovsdb
+const OvsdbConnected = "connected"
+const OvsdbRemoved = "removed"
 
 // ovsdbSpec is a helper structure for precise identification of ovsdb databases. A lot of
 // ovn/ovs commands take path to either database file, database socket or process control socket
@@ -59,11 +61,11 @@ func newOvsdbSpec(dbType OvsdbType) (*ovsdbSpec, error) {
 	return dbSpec, err
 }
 
-// waitForDBConnected as the name suggests, waits for specified ovsdb database to settle in
-// "connected" state. If database does not reach this state within timeout, this function returns error.
+// waitForDBState as the name suggests, waits for specified ovsdb database to settle in
+// specified state. If database does not reach this state within timeout, this function returns error.
 // Target specified in "db" parameter does not need to necessarily exist before this function is executed,
 // creation of the database socket (db.Target) will be awaited as well.
-func waitForDBConnected(s *state.State, db *ovsdbSpec, timeout int) error {
+func waitForDBState(s *state.State, db *ovsdbSpec, dbState string, timeout int) error {
 	_, err := shared.RunCommandContext(
 		s.Context,
 		"ovsdb-client",
@@ -72,10 +74,10 @@ func waitForDBConnected(s *state.State, db *ovsdbSpec, timeout int) error {
 		"wait",
 		fmt.Sprintf("unix:%s", db.Target),
 		db.Name,
-		"connected",
+		dbState,
 	)
 	if err != nil {
-		return fmt.Errorf("failed to connect to %s database in '%s': %w", db.Name, db.Target, err)
+		return fmt.Errorf("database in '%s' (%s) failed to reach state '%s': %w", db.Name, db.Target, dbState, err)
 	}
 	return nil
 }
@@ -101,7 +103,7 @@ func ovnDBCtl(s *state.State, dbType OvsdbType, timeout int, args ...string) (st
 		return "", err
 	}
 
-	err = waitForDBConnected(s, dbSpec, timeout)
+	err = waitForDBState(s, dbSpec, OvsdbConnected, timeout)
 	if err != nil {
 		return "", err
 	}
@@ -136,7 +138,7 @@ func VSCtl(s *state.State, args ...string) (string, error) {
 		return "", err
 	}
 
-	err = waitForDBConnected(s, dbSpec, defaultDBConnectWait)
+	err = waitForDBState(s, dbSpec, OvsdbConnected, defaultDBConnectWait)
 	if err != nil {
 		return "", err
 	}
--- a/microovn/ovn/commands.go
+++ b/microovn/ovn/commands.go
@@ -3,6 +3,7 @@ package ovn
 import (
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
 
 	"github.com/canonical/microcluster/state"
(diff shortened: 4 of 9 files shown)


microovn/ovn/certificates.go
mkalcok, line 1: This is not connected to the rest of the change. I just noticed that the returned error was not propagated

microovn/ovn/commands.go
```diff
@@ -59,11 +62,11 @@ func newOvsdbSpec(dbType OvsdbType) (*ovsdbSpec, error) {
 	return dbSpec, err
 }
```
mkalcok, line 64: Function below was changed to be more flexible and allow us to wait on arbitrary database state. Not just `connected`

microovn/ovn/commands.go
```diff
@@ -136,14 +139,46 @@ func VSCtl(s *state.State, args ...string) (string, error) {
...
+func AppCtl(s *state.State, target string, args ...string) (string, error) {
+	arguments := []string{"-t", target}
+	arguments = append(arguments, args...)
+	return shared.RunCommandContext(
+		s.Context,
+		"ovn-appctl",
+		arguments...,
+	)
+}
+
```
mkalcok, line 162: `ovn-controller` uses a a control socket with random number in the filename (i.e. `ovn-controller.3376923.ctl=`) unlike OVN SB or NB processes which use predictable names like `ovnnb_db.ctl=`. However `ovs-appctl` can figure out correct control socket used by `ovn-controller` if env variable `OVS_RUNDIR` is properly set. :point_down:

microovn/database/schema.go
```diff
@@ -36,3 +37,27 @@ CREATE TABLE services (
 
 	return err
 }
+
```
mkalcok, line 40: Sorry for tagging you directly @masnax but I'm having an issue with this schema migration (independently of any other change in this PR) and you were very knowledgeable about the `microcluster` last time we spoke, so  I was hoping you could maybe point me in a right direction.
About 1/10th of a bootstrap attempts fail with
```
Error: failed to apply update 1: no such table: services
```
I'd hate to jump to conclusions, but I don't see how this error could originate just from the statements in the new transaction. Is it possible that there's some issue with the way `microcluster` executes schema updates so that the new update (below) is executed before the first update (above) properly finishes?
(@fnordahl just 'cause we talked about this yesterday)

mkalcok: You are the best! Thank you very much

microovn/ovn/certificates.go
mkalcok, line 1: This is not connected to the rest of the change. I just noticed that the returned error was not propagated

microovn/ovn/commands.go
fnordahl: The OVN project provides its own appctl variant, `ovn-appctl`, which will have the appropriate defaults to handle this communication without the extra arguments.

So can't you just use the above `AppCtl` function for this, and if not, there may be environment variables not set up properly?

Example from a classic deployment:
```
$ sudo ovn-appctl list-commands
The available commands are:
  connection-status       
  coverage/read-counter   COUNTER
  coverage/show           
  ct-zone-list            
...
$ sudo ovn-appctl -t ovn-controller list-commands
The available commands are:
  connection-status       
  coverage/read-counter   COUNTER
  coverage/show           
  ct-zone-list            
```

microovn/ovn/environment.go
```diff
@@ -179,3 +179,17 @@ func createPaths() error {
 
 	return nil
 }
+
+// removePaths removes directories created by createPaths function. This effectively removes
+// any data created during MicroOVN runtime.
+func removePaths() []error {
```
fnordahl, line 185: While I like the way you think to be a good citizen and clean up after ourself, I think this should be a conscious choice of the operator, and not something we do automatically.

Would not calling `snap remove --purge` accomplish the same thing for the operator?

fnordahl (changes requested):
Thank you for your work on this, Martin, and for your persistence in getting issues resolved in our dependencies for the schema migration.

I have a couple of comments in-line.

I wanted to test, but it was not clear to me exactly what the user exposed process would be, does this deserve some documentation? It also highlights a need for us to ramp up our game in functional testing for microovn, not saying you need to implement the required scaffolding as part of this change, but we need to think about how/when we get that in place.

microovn/ovn/commands.go
mkalcok: `ovn-appctl` and `ovs-appctl` both behave exactly the same in regards to the "ovn-controller". You can specify `-t controller` for both of them and the auto-discovery mechanism will work as long as runtime directory is properly set.
* `$OVS_RUNDIR` for `ovs-appctl`
* `$OVN_RUNDIR` for `ovn-appctl`

I chose to use `ovs-appctl` just because that's what's mentioned in [the docs](https://man7.org/linux/man-pages/man8/ovn-controller.8.html) but I also tested it and both commands work the same. At the end of the day, we need to set one of these two env vars and invoke matching command with `-t controller`.

I also though about adding new wrapper that would set "rundir" variable to chassis' runtime directory but that seemed like a bigger disruption to the UX that would have to be properly documented and maintained.

fnordahl: > `ovn-appctl` and `ovs-appctl` both behave exactly the same in regards to the "ovn-controller". You can specify `-t controller` for both of them and the auto-discovery mechanism will work as long as runtime directory is properly set.
> 
>     * `$OVS_RUNDIR` for `ovs-appctl`
> 
>     * `$OVN_RUNDIR` for `ovn-appctl`
> 
> 
> I chose to use `ovs-appctl` just because that's what's mentioned in [the docs](https://man7.org/linux/man-pages/man8/ovn-controller.8.html) but I also tested it and both commands work the same. At the end of the day, we need to set one of these two env vars and invoke matching command with `-t controller`.

The `ovs-appctl` in the docs is much likely a left-over from OVS/OVN splitting into separate projects.

When not having these tools in a snap, the build time configuration would make the tool dtrt. on its own, so the environment variable is a snap problem.

> I also though about adding new wrapper that would set "rundir" variable to chassis' runtime directory but that seemed like a bigger disruption to the UX that would have to be properly documented and maintained.

Which UX are you thinking of here developer UX, the `Appctl` function is not exposed to the CLI right?

I think it would be more appropriate to have `OvsAppCtl` and `OvnAppCtl` functions that set the environment variable appropriately. These tools have a wide range of uses and we may need generic access to them later, rater than having a super specialized `ControllerCtl`.

microovn/ovn/environment.go
mkalcok: Even the regular `snap remove` will accomplish it. The `--purge` just prevents creation of a snapshot.

I decided to remove all data files to prevent unintended ovsdb cluster re-connection. I'm going to double-test this, but I'm pretty sure that if we don't delete the files on departure and someone manually restarts services with `snap start microovn.central`, they'd try to re-join the cluster. This would create inconsistent state in which member would show up in `ovn-nbctl cluster/status` but it would not be present in `microovn cluster list`.

I agree that perhaps we can be less drastic with the cleanup and we could definitely keep at least `logs/` around. But I would for sure wipe at least `ovn.env` file and certificates. Do you see any value in leaving database files around?

I'd definitely have to re-test workflow
* join member
* kick member
* re-join member

which works nicely right now.

microovn/ovn/commands.go
mkalcok: >> I also though about adding new wrapper that would set "rundir" variable to chassis' runtime directory but that seemed like a bigger disruption to the UX that would have to be properly documented and maintained.

> Which UX are you thinking of here, developer UX? The Appctl function is not exposed to the CLI right?

I thought about creating the snap/cli wrapper, like `microovn.controller-ctl`, but I decided against it.

> I think it would be more appropriate to have OvsAppCtl and OvnAppCtl functions that set the environment variable appropriately. These tools have a wide range of uses and we may need generic access to them later, rather than having a super specialized ControllerCtl.

I agree with you that we should have general `OvsAppCtl` and `OvnAppCtl` However if we were to set runtime directories for these commands automatically we'd set
* `$SNAP_COMMON/run/central` for `OvnAppCtl`
* `$SNAP_COMMON/run/switch` for `OvsAppCtl`
But neither of these is appropriate when we want to target controller which needs `$SNAP_COMMON/run/chassis`. That's why I thought that adding the third function would be helpful. We can also make the wrappers really bare-bone and force user to pass all required arguments but I think that detracts a bit from usefulness.

mkalcok: Sorry. I kinda forgot to conclude my thought in previous comment. My proposal is that we have three functions
* Generic `OvnAppCtl`
* Generic `OvsAppCtl` (there wasn't a use case for this one yet so it doesn't exist)
* "Helper" `ControllerCtl` that uses either `ovs-appctl` or `ovn-appctl` and sets proper runtime directory for targeting controller.

What do you think about this approach?

fnordahl: > Sorry. I kinda forgot to conclude my thought in previous comment. My proposal is that we have three functions
> 
>     * Generic `OvnAppCtl`
> 
>     * Generic `OvsAppCtl` (there wasn't a use case for this one yet so it doesn't exist)

:+1: 
 
>     * "Helper" `ControllerCtl` that uses either `ovs-appctl` or `ovn-appctl` and sets proper runtime directory for targeting controller.

I think we should avoid this, and instead figure out why this does not work out of the box.

Having thought some more about your comment about UX, the end user of `microovn.ovs-appctl` and `microovn.ovn-appctl` would expect those tools to make the appropriate decisions by default. Why are they currently not doing so?

Looking at microovn/snapcraft/ovn.env sourced by the snap `ovn-appctl` script:
https://github.com/canonical/microovn/blob/407a98a32ff5e8c8e8e1e0ad5d933bb5c159f44f/snapcraft/ovn.env#L9

Perhaps that is causing the issue?

The ovs-* commands already set `OVS_RUNDIR` here:
https://github.com/canonical/microovn/blob/407a98a32ff5e8c8e8e1e0ad5d933bb5c159f44f/snapcraft/commands/ovs#L2

mkalcok: >  export OVN_RUNDIR="${SNAP_COMMON}/run/switch/"

This is indeed strange and looks like a typo?

> I think we should avoid this, and instead figure out why this does not work out of the box.

We have two CLI tools

* `ovn-appctl`
* `ovs-appctl`

And three runtime directories

`$SNAP_COMMON/run/central`
`$SNAP_COMMON/run/switch`
`$SNAP_COMMON/run/chassis`

So I think we have two options:

* Add 3rd wrapper (be it in code, or as a CLI wrapper/script) and each can default to correct runtime dir
* Don't add 3rd wrapper but make "runtime dir" a required argument for both `OvsAppCtl` and `OvnAppCtl`.

The magic `-t controller` works by looking for `ovn-controller.pid` file in `$OVS_RUNDIR` or `$OVN_RUNDIR` (based on whether you are running `ovs-appctl` or `ovn-appctl` respectively). It then takes the pid and looks up socket associated with that process.
So another alternative is to just cancel `$SNAP_COMMON/run/chassis/` and have controller pid and socket files in `$SNAP_COMMON/run/switch`. This way the `ovs-appctl` would work out of the box with `-t controller`. However that looked like a pretty big structure change to me so I didn't go for it.

fnordahl: Ooh, it did not occur to me that we had split the runtime directories across three paths, I see the predicament now.

OK, you're suggestion is then fine, lets go with it.

In a future change, I think we should merge `$SNAP_COMMON/run/central` and `$SNAP_COMMON/run/chassis` into `$SNAP_COMMON/run/ovn`, unless there are strong counter arguments against it. That would rid us of this problem and also retain compatibility with upstream intended use of these tools.

What do you think?

mkalcok: Yup, I fully agree. If there are no blockers for merging of runtime dirs, we should do it.

microovn/ovn/environment.go
mkalcok: As per our discussion. Instead of fully removing the directory, we can move & timestamp data directory. I'll make the change.

microovn/ovn/environment.go
```diff
@@ -179,3 +182,55 @@ func createPaths() error {
 
 	return nil
 }
+
+// cleanupPaths backs up directories defined by paths.BackupDirs and then removes directories
+// created by createPaths function. This effectively removes any data created during MicroOVN runtime.
+func cleanupPaths() []error {
```
fnordahl, line 188: Returning errors as a `Slice` comes across as a bit odd to me, looking at the code below, it will eventually bail on error before completing processing anyway, so why not just bail on the first error with a regular error return?

In one way or the other the system will be left in an inconsistent state, and it is very likely that the cause of the first error will be the same as the cause of subsequent errors.

Had there been some alternative strategy that could resolve any of the errors, I would have understood it better, but I can't see any way to predict how to deal with failure here?

microovn/ovn/environment.go
```diff
@@ -179,3 +182,55 @@ func createPaths() error {
...
+		if err != nil {
+			errors = append(errors, err)
+		}
+	}
+
+	// Return if any backups failed
+	if len(errors) > 0 {
+		errors = append(
+			errors,
+			fmt.Errorf("failures occured furing backup. Refusing to continue with data removal"),
```
fnordahl, line 221: Following on the comment at the begining of the function, I'd move this message to where `cleanupPaths` is called after bailing from this function on the first error.

README.md
```diff
@@ -81,6 +81,49 @@ microovn.ovn-sbctl show
...
+
+MicroOVN will attempt to gracefully remove services running on the removed
+member from the OVN cluster. You can watch logs on departing member for any
+indications of removal failure with:
+
+```shell
+snap logs -f microovn.daemon
+```
+
+If there are any issues during the removal process, you will have to take
```
fnordahl, line 100: Do we have known issues for the removal process, if so should it be listed? If not, do we need to mention this?

README.md
```diff
@@ -81,6 +81,49 @@ microovn.ovn-sbctl show
...
 
+## Downscaling
+
+Removing MicroOVN cluster member is as easy as running
+
+```shell
+microovn cluster remove <member_name>
+```
+
+MicroOVN will attempt to gracefully remove services running on the removed
```
fnordahl, line 92: the grammar here is a bit off, because as the removal process is in progress, a member is not yet removed.

What about something like:
MicroOVN will stop and disable any chassis components running on the member (`ovn-controller` and `ovs-vswitchd`), before attempting to gracefully leave any OVN database clusters for members with central components present.

fnordahl (changes requested):
Thank you for your work on this, Martin. I took it for a spin and it works great!

I have a few comments in-line.

For a future iteration we should probably look into integration of this with a snap remove hook.

README.md
```diff
@@ -81,6 +81,49 @@ microovn.ovn-sbctl show
 microovn.ovn-appctl -t /var/snap/microovn/common/run/central/ovnnb_db.ctl cluster/status OVN_Northbound
 ```
 
+## Downscaling
```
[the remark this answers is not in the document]
mkalcok: I think we need to balance our documentation so that we won't end up copy-pasting upstream project documentation details. Perhaps rather than going into details about Raft and OVN resiliency, I could find relevant upstream documentation and link it?

microovn/ovn/environment.go
mkalcok: But errors can happen for other reasons than failed backup. If we logged this message outside of a context of this function, we'd have to somehow pass error "reason" to the caller.

README.md
mkalcok: Of the top off my head I can think of:
* Service already stopped on the node that we are removing
* Network connectivity issues between OVN services
* Someone removed certificates on the node that we are removing
* Someone messed up configuration of ovn/microovn on the node that we are removing

I think there may be many problems, often outside of MicroOVN control. That's why I opted for vague wording here, and pointed user to the logs which will tell them what exactly failed during removal process.

README.md
mkalcok: This sounds much better. Thank you!

microovn/ovn/environment.go
mkalcok: While the errors during backup cause this function to bail with single error (to prevent deletion). The part that is attempting to delete files (after backup), compounds errors without bailing, so that we can later log lists of directories that we failed to delete.
Now this is a minor functionality and could be argued that it's not that important. I can change it if you still don't like the slices.

tomponline: FWIW if you do stick with multiple errors, you can use errors.Join() from the stdlib which might make it more consumable:

https://pkg.go.dev/errors#Join

README.md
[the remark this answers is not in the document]
fnordahl: Sure, but I think it is important to highlight that two members with central components is worse than having one, so that people don't leave their clusters rubbing like that for extended periods of time.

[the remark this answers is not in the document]
mkalcok: Yup, definitely +1. I'll try to find some good source on that to link. Or if I can't find any I'll just add the warning explicitly.

fnordahl (approved):
LGTM, thanks!

PR #38, opened 2023-06-23
microovn/ovn/environment: Use dqlite cluster record to find peers

masnax:
Rather than using `state.Remotes` which is the local record of nodes and is updated on each heartbeat, use `cluster.InternalClusterMember` to get the most up-to-date record of peers.

The `services` table is already related to the `internal_cluster_members` table so this will ensure every peer is picked up for every service.

### DIFF
--- a/microovn/ovn/environment.go
+++ b/microovn/ovn/environment.go
@@ -5,14 +5,17 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"net"
 	"net/netip"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
 
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/microcluster/cluster"
 	"github.com/canonical/microcluster/state"
 
 	"github.com/canonical/microovn/microovn/database"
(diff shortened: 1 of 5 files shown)


microovn/ovn/environment.go
```diff
@@ -121,15 +134,23 @@ func generateEnvironment(s *state.State) error {
 		}
 
 		server := servers[0]
```
mkalcok, line 136: It looks to me that there's a quite bit of code overlap between this part of code that generates `{n,s}bInitial` and the code in `connectString()`. What would you think about returning  "initial string" as part of the `connectString()`? Perhaps by changing return value from `string` to some simple `struct` with 2 members?
If not in this PR, maybe we can just open an issue about it.

masnax: Looking at it, you're right, there's quite a bit of duplicated code and a redundant db transaction as well.

I've went ahead and renamed `connectString` to `environmentString` and modified it to fetch all the data from one transaction, and parse both the connect strings and initial strings afterward. 

I opted to just have the function return two strings for now rather than build out a whole struct, that might be something to consider if this helper continues to grow in scope, or if it ends up that this information gets passed around together often.

mkalcok (approved):
Not sure if this needs another pair of eyes but it does LGTM. Thanks :)

PR #40, opened 2023-06-30
Add docs directory for RTD - again

pmatulis:
This PR adds a `docs` directory to the root of the repository. This is to have MicroOVN documentation published on readthedocs.com (RTD). The origin of this docs setup is the [sphinx-docs-starter-pack](https://github.com/canonical/sphinx-docs-starter-pack) project, which is what all RTD-based company projects are expected to use.

As agreed, all the stub pages have been removed. The landing page has just a few notes on next steps.

Doc versioning is possible but is currently disabled.

There is a link enabled in the footer for asking community questions in https://discourse.ubuntu.com/ but I recommend that we soon carve out a separate space (Discourse category) for MicroOVN on that forum - and change the link when that's done.


docs/.wordlist.txt
mkalcok, line 1: If this is a list of words for autocorrect, I'd suggest following
* ovsdb
* OVN
* ovnnb
* ovnsb
* northd
* openvswitch

docs/.github/workflows/automatic-doc-checks.yml
mkalcok, line 1: I'd suggest removing this directory entirely. We have copy of this workflow file in root workflow directory (where it should be imo) and this additional copy could cause some confusion.

mkalcok (changes requested):
Thank you for jump starting docs Peter. Just two minor suggestions below. Otherwise it LGTM.

docs/.github/workflows/automatic-doc-checks.yml
pmatulis: Originally I wanted to use a symlink to the docs directory but I read that this wouldn't work with GitHub. I'll remove it.

docs/.wordlist.txt
pmatulis: I'll add these words.

mkalcok (approved):
LGTM. Thank you.

PR #43, opened 2023-07-05
Fix doc build

pmatulis:
The doc build was failing due to a documentation resource file (`doc-cheat-sheet`) that was actively referring to the doc `readme.rst` file. The latter was hollowed out in a previous PR, causing the cheat sheet to complain. The latter will be redone (in its own repository) to make it passive.

The `readme.rst` file was also orphaned in order to remove a doc build warning.


fnordahl (approved):
LGTM, thanks!

PR #45, opened 2023-07-07
[docs] Fix more build errors

pmatulis:
Some sub-optimally represented doc content is still causing build errors.

Remove a cheat sheet for MyST (a Markdown implementation that will never be used in the current repository) and make a section of the RST cheat sheet passive instead of active (by putting it in a code block).


mkalcok (approved):
Thanks for the fixes. LGTM

PR #49, opened 2023-07-26
snap: Switch main branch to core24/devel build-base

fnordahl:

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -214,8 +214,7 @@ parts:
      - bin/ovsdb-client
      - bin/ovsdb-server
      - bin/ovsdb-tool
-     - lib/*/libevent-2.1.so*
-     - lib/*/libunbound.so*
+     - lib/*
      - share/openvswitch
     override-build: |
         craftctl default
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,10 +1,11 @@
 name: microovn
 icon: microovn.png
-base: core22
+base: core24
+build-base: devel
 assumes:
  - snapd2.59
 adopt-info: ovn
-grade: stable
+grade: devel
 source-code: https://github.com/canonical/microovn.git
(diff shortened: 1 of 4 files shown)


snap/snapcraft.yaml
```diff
@@ -1,10 +1,11 @@
 name: microovn
 icon: microovn.png
-base: core22
+base: core24
+build-base: devel
 assumes:
  - snapd2.59
 adopt-info: ovn
-grade: stable
+grade: devel
```
mkalcok, line 8: I wonder what this means for our release process. `devel` snaps can't go into `candidate` (and above) channels. So, for any new release we'd cut the branch and change this to `stable`?

mkalcok (approved):
One in-line question, otherwise LGTM

snap/snapcraft.yaml
fnordahl: Presumably, yes. We already need to lock down certain pieces in the `snapcraft.yaml` as part of the release process.

To maintain end to end integrity in the git history, we can also momentarily change the main branch to what the stable branch will become. Upstream follows a process like this.

Example from the most recent OVS release, main branch:
https://github.com/openvswitch/ovs/commit/f20980a19eb57de24aef29d1d4d5b80b61a9a982
https://github.com/openvswitch/ovs/commit/bffffd841ff1a201fcf5bd7007d64e1d511d6aa1
https://github.com/openvswitch/ovs/commit/de86c5bbdc627117ce8a31f9c9a599c372c8e2a8

branch-3.2 is then created from https://github.com/openvswitch/ovs/commit/f20980a19eb57de24aef29d1d4d5b80b61a9a982

The commit setting the date is cherry picked: https://github.com/openvswitch/ovs/commit/f8d24583ed53d54ddb78c90d5b53103132447b98
And then preparing for next point release is of course deviating from the main branch: https://github.com/openvswitch/ovs/commit/e99e98655b7f1a6d8911d97430adc7056c67a8d0

PR #51, opened 2023-07-31
[WIP] test: Add autotest framework for functional tests

fnordahl:
One of the possible frameworks we may use for our testing is the autotest framework.

This patch put together a minimal set of macros and tests to demonstrate how it could be used with LXD as a backend for running the contained code.

While it is a bit archaic, it is very simple and self contained, it is also well understood by the pool of people interested in working on OVS and OVN.

PR #52, opened 2023-07-31
[WIP]tests: Setup for functional tests using BATS

mkalcok:
Alternative approach to #51 using [`bats` testing framework](https://bats-core.readthedocs.io/en/stable/index.html). Attached is also a `Makefile` that outlines how the tests would be executed. For some reason, the tests fail when executed via Makefile but pass when executed manually, but that's not that important right now.

Pros of `bats`:
* Provides concept of [per-file](https://bats-core.readthedocs.io/en/stable/tutorial.html#avoiding-costly-repeated-setups) and [per-test](https://bats-core.readthedocs.io/en/stable/tutorial.html#let-s-do-some-setup) `setup`s and `teardown`s
* allows easily sharing code via [`load`](https://bats-core.readthedocs.io/en/stable/writing-tests.html#load-share-common-code) functionality 
* Allows parallel execution
* Tests are written in almost pure bash with [an option to write completely pure bash](https://bats-core.readthedocs.io/en/stable/faq.html#how-can-i-lint-shell-format-my-bats-tests) 

Cons of `bats`:
* While base package is available via `apt` addons providing basic functionality and QOL (like asserts) need to be installed manually as git submodule (or via `npm`)
* Failures output/reason is not always super clear
* We still need to write a lot of helpers that deal with execution of code in LXC containers ourselves

Example output (currently containing a lot of debug prints):
```
$ bats tests/basic_cluster.bats 
basic_cluster.bats
   Launching microovn-test-0
   Launching microovn-test-1
   Launching microovn-test-2
   Deploying MicroOVN to microovn-test-0
   Waiting for snapd to be ready on microovn-test-0
   Installing MicroOVN in container microovn-test-0
   Deploying MicroOVN to microovn-test-1
   Waiting for snapd to be ready on microovn-test-1
   Installing MicroOVN in container microovn-test-1
   Deploying MicroOVN to microovn-test-2
   Waiting for snapd to be ready on microovn-test-2
   Installing MicroOVN in container microovn-test-2
   Bootstrapping MicroOVN on microovn-test-0
   Adding microovn-test-1 to the cluster
   Joining cluster with microovn-test-1
   Adding microovn-test-2 to the cluster
   Joining cluster with microovn-test-2
 ✓ Expected MicroOVN cluster count
 ✓ Expected services up
   Cleaning up microovn-test-0
   Cleaning up microovn-test-1
   Cleaning up microovn-test-2

2 tests, 0 failures
```


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "tests/test_helper/bats-support"]
+	path = tests/test_helper/bats-support
+	url = https://github.com/bats-core/bats-support.git
```
fnordahl, line 3: While I at first glance thought the submodule approach was a bit odd for a test dependency, I do prefer it over having to pollute the system running the test with `npm install`, so ++

I also see that these are available as packages in recent Ubuntu releases, so come next LTS we may no longer need to do this:
https://launchpad.net/ubuntu/+source/bats-assert
https://launchpad.net/ubuntu/+source/bats-support

tests/test_helper/lxd.bash
```diff
@@ -0,0 +1,40 @@
+LXC_TEST_IMAGE=jammy
+DEPENDENCIES="jq"
+
+function start_containers() {
+    LXC_COUNT=$1
+    LXC_BASE_NAME="microovn-test"
+    ALL_CONTAINERS=""
+    for (( i = 0; i < LXC_COUNT; i++ )); do
+        CONTAINER_NAME="$LXC_BASE_NAME-$i"
```
fnordahl, line 9: If we run multiple tests in parallel, they would most likely step on each other with this naming.

Does bats provide some variable identifying each test that we could use?

Makefile
```diff
@@ -0,0 +1,8 @@
+MICROOVN_SNAP_PATH=$(CURDIR)/microovn.snap
+build:
+	@echo "Building the snap"
+	@snapcraft pack --use-lxd -o $MICROOVN_SNAP_PATH
+
+func-tests:
```
fnordahl, line 6: `func-tests` does not ring too well with me, maybe use a target name starting with `check` to be in line with our upstreams?

Add the .snap file as a dependency so that the build is done automatically?
```make
check-system: $(MICROOVN_SNAP_PATH)
```

Makefile
```diff
@@ -0,0 +1,8 @@
...
+	@echo "Running functional tests"
+	@bats tests/basic_cluster.bats
```
fnordahl, line 8: I believe the target provided to `bats` can be a directory so that it will execute all the *.bats file in that directory.

tests/test_helper/lxd.bash
```diff
@@ -0,0 +1,40 @@
...
+        echo "# Launching $CONTAINER_NAME" >&3
+        lxc launch -q ubuntu:$LXC_TEST_IMAGE $CONTAINER_NAME
+        lxc_exec $CONTAINER_NAME "apt update && apt install -y $DEPENDENCIES"
+        ALL_CONTAINERS+="$CONTAINER_NAME "
+    done
+
+    export ALL_CONTAINERS
```
fnordahl, line 16: We should be very careful with global variables in library/utility code. I wonder if it would be as easy for the calling code to provide identifiers for instances to create or something similar?

tests/test_helper/lxd.bash
```diff
@@ -0,0 +1,40 @@
+LXC_TEST_IMAGE=jammy
+DEPENDENCIES="jq"
+
+function start_containers() {
```
fnordahl, line 4: The functions in this file can potentially be used with multiple backends which we could possibly implement in other modules, for the LXD backend it can also be used for both containers and virtual machines.

I wonder if we should use a more generic name, such as `instances`, so that we can use it for both? We then have an opportunity to run the same tests using containers and virtual machines by modifying an environment variable or something similar.

tests/test_helper/lxd.bash
```diff
@@ -0,0 +1,40 @@
...
+
+function cleanup_containers() {
+    IFS=" "
+    for container in $ALL_CONTAINERS ; do
+        echo "# Cleaning up $container" >&3
+        lxc delete --force $container
+    done
+}
+
+function lxc_exec() {
```
fnordahl, line 27: To support a possible future of multiple backends, should we omit `lxd` in the name here and just do something like `instance_exec`? I guess a shorter name would be better though.

fnordahl (commented):
I like the proposed approach, and have to admit I don't mind not having to gripe with m4. Having the test themselves in regular bash code makes it easy to transport to anything else if we ever change our minds.

Added some comments that came to mind as I read through this.

> Cons of bats:
>
> * While base package is available via apt addons providing basic functionality and QOL (like asserts) need to be installed manually as git submodule (or via npm)

It looks like these addons are finding their way into Ubuntu, and the submodule proposal works just fine!

> * Failures output/reason is not always super clear

I can't think of any framework that does this for you, I guess we just need to be rigorous in our tests and ensure required logs are provided along with test results.

> * We still need to write a lot of helpers that deal with execution of code in LXC containers ourselves

There is no way around this, other alternatives include ansible playbooks or terraform plans but that just adds more dependencies for the test. I think our needs for this are generic and simple enough to approach the way you propose here.

fnordahl:
To unblock further development I moved this branch to canonical/microovn/bats-testing and #53

PR #54, opened 2023-08-02
tests/basic_cluster: Fix sharing of TEST_CONTAINERS

fnordahl:
The committed solution for sharing TEST_CONTAINERS between file level setup and teardown functions and the individual tests does not actually work.

This was not caught because the individual tests do their assertions within loops, they provided false positive results when the loop variable was empty.

Add assertion on contents of TEST_CONTAINERS in per-test setup function to avoid false positive results.

Export the TEST_CONTAINERS variable from the file-level setup function.  This is in line with the bats documentation and have confirmed it works fine even when running multiple test files in parallel.

Add a shell fragment comment at the top so that editors/IDEs enable bash syntax highlighting when editing the file.

### DIFF
--- a/tests/basic_cluster.bats
+++ b/tests/basic_cluster.bats
@@ -1,4 +1,4 @@
-TEST_CONTAINERS=""
+# This is a bash shell fragment -*- bash -*-
 
 setup_file() {
     load test_helper/common.bash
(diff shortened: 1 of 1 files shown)


mkalcok (approved):
Thanks for catching this tricky bug

PR #55, opened 2023-08-02
[WIP] tests: Add init_cluster and init_cluster_ipv6 tests

fnordahl:
These tests are similar to the basic_cluster test, the main difference being that the cluster is initialized using the `microovn init` command.

This serves multiple purposes:

* Check behavior of the `microovn init` command.

* Check clustering behavior when user supplies IPv4 address.

* Check clustering behavior when user supplies IPv6 address.

TODO:

* Figure out organization of shared setup / test code.
  * In this iteration I deliberately copied most of `basic_cluster.bats` over to `init_cluster.bats`. Need to evaluate if all the tests are relevant for both cases, and if so figure out how to avoid the copy/paste.
* Compare and contrast the use of a symlink to create a test matrix for the `init_cluster` to other approaches.
  * Pro: The symlink approach works quite neatly in that it allows for parallel execution both when the `bats` command handles parallelization, and when something like GHA handles parallelization (presumably through test matrix environment variables and execution of multiple copies of `bats` pointed directly at individual files).
  * Pro: The test matrix expansion is also implemented without adding any boiler plate or cargo culting stuff around.
  * Con: Its implied nature may not be obvious though and there would also be benefits to a more explicit approach.
  * Let's find out if this way is acceptable or if we find a more explicit way to compose this without boiler plate and copy paste.


mkalcok (approved):
LGTM

PR #56, opened 2023-08-03
tests: Parallelize container creation

fnordahl:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,11 @@ check-tabs:
 	grep -lrP "\t" tests/ && exit 1 || exit 0
 
 check-lint: check-tabs
-	find tests/ -type f | xargs shellcheck && echo Success!
+	find tests/ \
+		-type f \
+		-not -name \*.yaml \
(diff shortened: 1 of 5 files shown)


mkalcok (commented):
Looks like a neat approach to me. I just wonder if we could maybe extract the cloudinit data into a separate file. Passing the "raw" yaml inline can get ugly, especially if  the number of dependencies grows. Also the file could be shared among multiple test suits. What do you think?

fnordahl:
> Looks like a neat approach to me. I just wonder if we could maybe extract the cloudinit data into a separate file. Passing the "raw" yaml inline can get ugly, especially if the number of dependencies grows. Also the file could be shared among multiple test suits. What do you think?

More files to manage is also not great. One cloud-init file per .bats file could perhaps work, but there is no guarantee that each .bats file manages only one type of containers sharing the same dependencies?

mkalcok (approved):
Thanks, LGTM. I'll watch for CI pass and I'll merge it

PR #57, opened 2023-08-03
tests/test_helper/lxd: Load openvswitch kernel module through LXD config

fnordahl:
To improve developer ergonomics, we ensure the `openvswitch` kernel module is loaded by making use of the
`linux.kernel_modules` LXD instance configuration option.

Doing it this way gives one less step for anyone wanting to run the test suite on their own system.

This makes it obvious that the `tests/cloud-init.yaml` file is badly named, so rename it.


mkalcok (approved):
LGTM

PR #58, opened 2023-08-03
Functional test for TLS

mkalcok:
This test suit checks that TLS functions properly in MicroOVN deployment.

This PR also includes "library" for in-test process parallelization that allows to start multiple functions/commands in background and later collect their results. The only caveat of using this parallelization that I discovered so far:

- it's not capable of collecting STDOUT/ERR output from successful background tasks (although in case of failure the output is printed out).
- Error output in case of failure is mixed. Due to the async execution, individual lines from multiple processes are "interwoven". This is not ideal but it's also not the worst imo as the cause of error can still be identified.

### DIFF
--- a/tests/test_helper/microovn.bash
+++ b/tests/test_helper/microovn.bash
@@ -44,8 +44,29 @@ function bootstrap_cluster() {
         echo "# Joining cluster with $container" >&3
         lxc_exec "$container" "microovn cluster join $token"
     done
+    wait_ovn_services $containers
 }
 
+# wait_ovn_services CONTAINER1 [CONTAINER2 ...]
+#
+# Wait for OVN "central" services running on the CONTAINERs to start listening.
+# Containers with disabled "central" service are skipped.
+function wait_ovn_services() {
+    local containers=$*
+    local ovn_ports="6641 6642 6643 6644"
+
+    for container in $containers; do
+        container_services=$(microovn_get_cluster_services "$container")
+        if [[ "$container_services" != *"central"* ]]; then
+            continue
+        fi
+
+        echo "# ($container) Waiting for OVN Central services to start"
+        for port in $ovn_ports; do
+            wait_for_open_port "$container" "$port" 30
+        done
+    done
+}
 function microovn_init_create_cluster() {
     local container=$1; shift
     local address=$1; shift
--- a/tests/test_helper/setup_teardown/tls_cluster.bash
+++ b/tests/test_helper/setup_teardown/tls_cluster.bash
@@ -1,0 +1,32 @@
+setup_file() {
+    load test_helper/common.bash
+    load test_helper/lxd.bash
+    load test_helper/microovn.bash
+
+
+    TEST_CONTAINERS=$(container_names "$BATS_TEST_FILENAME" 4)
+    CENTRAL_CONTAINERS=""
+    CHASSIS_CONTAINERS=""
+
+    export TEST_CONTAINERS
+    export CENTRAL_CONTAINERS
+    export CHASSIS_CONTAINERS
+    launch_containers jammy $TEST_CONTAINERS
+    install_microovn "$MICROOVN_SNAP_PATH" $TEST_CONTAINERS
+    bootstrap_cluster $TEST_CONTAINERS
+
+    # Categorize containers as "CENTRAL" and "CHASSIS" based on the services they run
+    for container in $TEST_CONTAINERS; do
+        container_services=$(microovn_get_cluster_services "$container")
+        if [[ "$container_services" == *"central"* ]]; then
+            CENTRAL_CONTAINERS+="$container "
+        else
+            CHASSIS_CONTAINERS+="$container "
+        fi
+    done
+
+}
+
+teardown_file() {
+    delete_containers $TEST_CONTAINERS
+}
--- a/tests/test_helper/tls.bash
+++ b/tests/test_helper/tls.bash
@@ -1,0 +1,121 @@
+_PKI_DIR="/var/snap/microovn/common/data/pki/"
+CA_CERT_PATH="$_PKI_DIR""cacert.pem"
+CLIENT_CERT_PATH="$_PKI_DIR""client-cert.pem"
+CLIENT_KEY_PATH="$_PKI_DIR""client-privkey.pem"
+CONTROLLER_CERT_PATH="$_PKI_DIR""ovn-controller-cert.pem"
+CONTROLLER_KEY_PATH="$_PKI_DIR""ovn-controller-privkey.pem"
+OVN_NB_CERT_PATH="$_PKI_DIR""ovnnb-cert.pem"
+OVN_NB_KEY_PATH="$_PKI_DIR""ovnnb-privkey.pem"
+OVN_SB_CERT_PATH="$_PKI_DIR""ovnsb-cert.pem"
+OVN_SB_KEY_PATH="$_PKI_DIR""ovnsb-privkey.pem"
+NORTHD_CERT_PATH="$_PKI_DIR""ovn-northd-cert.pem"
+NORTHD_KEY_PATH="$_PKI_DIR""ovn-northd-privkey.pem"
+
+# shellcheck disable=SC2034  # This variable is referenced and passed by name
+declare -g -A OVN_CENTRAL_PKI=(\
+    [$CLIENT_CERT_PATH]=$CLIENT_KEY_PATH\
+    [$CONTROLLER_CERT_PATH]=$CONTROLLER_KEY_PATH\
+    [$OVN_NB_CERT_PATH]=$OVN_NB_KEY_PATH\
+    [$OVN_SB_CERT_PATH]=$OVN_SB_KEY_PATH\
+    [$NORTHD_CERT_PATH]=$NORTHD_KEY_PATH\
+)
+
+# shellcheck disable=SC2034  # This variable is referenced and passed by name
+declare -g -A OVN_CHASSIS_PKI=(\
+     [$CLIENT_CERT_PATH]=$CLIENT_KEY_PATH\
+     [$CONTROLLER_CERT_PATH]=$CONTROLLER_KEY_PATH\
+ )
+
+# verify_service_cert CONTAINER IP_ADDR PORT
+#
+# Ensure that service listening on IP_ADDR and PORT uses TLS and can be verified using
+# CA certificate used by the MicroOVN on the specified CONTAINER.
+function verify_service_cert() {
+    local container=$1
+    local ip_addr=$2
+    local port=$3
+    echo "# ($container) Checking TLS on $ip_addr:$port"
+    lxc_exec "$container" "openssl s_client -CAfile $CA_CERT_PATH -cert $CLIENT_CERT_PATH \
+        -key $CLIENT_KEY_PATH -verify_return_error -connect $ip_addr:$port <<< Q"
+}
+
+# _verify_cert_files CONTAINER CERTIFICATE_MAP
+#
+# Verify certificate files and keys specified by CERTIFICATE_MAP stored in the CONTAINER.
+#
+# CERTIFICATE_MAP should be passed as a variable name (not as a direct value) and the variable
+# is expected to be an associative array that maps local paths to certificate to local paths of
+# private keys. (see OVN_CENTRAL_PKI for example)
+#
+# Following checks are performed by this function:
+#   * Ensure that the private key is used to sign the certificate
+#   * Ensure that the certificate can be validated using CA certificate
+function _verify_cert_files() {
+    local container=$1
+    local -n certificate_map=$2
+    for cert in "${!certificate_map[@]}"; do
+        local key="${certificate_map[$cert]}"
+
+        echo "# ($container) Checking certificate $cert"
+        # Check that private key matches the certificate
+        cert_pubkey=$(lxc_exec "$container" "openssl x509 -noout -pubkey -in $cert")
+        key_pubkey=$(lxc_exec "$container" "openssl ec -pubout -in $key")
+        ## Check that relevant public key is found both in cert and in private key
+        assert [ -n "$cert_pubkey" ]
+        assert [ -n "$key_pubkey" ]
+        ## Ensure that public key hashes match
+        assert_equal "$(sha256sum "$cert_pubkey")" "$(sha256sum "$key_pubkey")"
+
+        # Check that certificates match CA
+        lxc_exec "$container" "openssl verify -CAfile $CA_CERT_PATH $cert"
+    done
+}
+
+# verify_central_cert_files CONTAINER
+#
+# Verify set of certificate files that are expected to be present on the CONTAINER
+# that runs "central" services.
+#
+# For more info about checks see _verify_cert_files
+function verify_central_cert_files() {
+    local container=$1
+    _verify_cert_files "$1" OVN_CENTRAL_PKI
+}
+
+# verify_central_cert_files CONTAINER
+#
+# Verify set of certificate files that are expected to be present on the CONTAINER
+# that runs "chassis" services.
+#
+# For more info about checks see _verify_cert_files
+function verify_chassis_cert_files() {
+    local container=$1
+    _verify_cert_files "$1"  OVN_CHASSIS_PKI
+}
+
+# reissue_certificate CONTAINER SERVICE
+#
+# Issue new certificate for a specified SERVICE that runs in the CONTAINER.
+#
+# This functions is a wrapper for executing 'microovn certificates reissue'. See help
+# output of that command for list of valid values for SERVICE argument.
+function reissue_certificate() {
+    # Issue new certificate for a specified OVN service in the container.
+    local container=$1
+    local service=$2
+    lxc_exec "$container" "microovn certificates reissue $service"
+}
+
+# get_cert_fingerprint CONTAINER CERT_PATH
+#
+# Print fingerprint of a certificate inside a CONTAINER
+function get_cert_fingerprint() {
+    local container=$1
+    local cert_path=$2
+    local fingerprint=""
+
+    fingerprint=$(lxc_exec "$container" "openssl x509 -in $cert_path -noout -fingerprint")
+    assert [ -n "$fingerprint"]
+
+    echo "$fingerprint"
+}
--- a/tests/test_helper/tls.bash
+++ b/tests/test_helper/tls.bash
@@ -31,9 +31,9 @@ declare -g -A OVN_CHASSIS_PKI=(\
 # Ensure that service listening on IP_ADDR and PORT uses TLS and can be verified using
 # CA certificate used by the MicroOVN on the specified CONTAINER.
 function verify_service_cert() {
-    local container=$1
-    local ip_addr=$2
-    local port=$3
+    local container=$1; shift
+    local ip_addr=$1; shift
+    local port=$1; shift
     echo "# ($container) Checking TLS on $ip_addr:$port"
     lxc_exec "$container" "openssl s_client -CAfile $CA_CERT_PATH -cert $CLIENT_CERT_PATH \
         -key $CLIENT_KEY_PATH -verify_return_error -connect $ip_addr:$port <<< Q"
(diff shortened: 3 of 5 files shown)


mkalcok:
re: failed tests
Seems like the first instance did not succesfully parse listening IP address..perhaps there's a race condition and I need to add some "wait" mechanism. I'm parsing the listening IP from `ss -ntlp | grep ovsdb-server | ....`

fnordahl:
> re: failed tests Seems like the first instance did not succesfully parse listening IP address..perhaps there's a race condition and I need to add some "wait" mechanism. I'm parsing the listening IP from `ss -ntlp | grep ovsdb-server | ....`

IIUC you parse the listening address because the daemon is configured to listen to all addresses.  You could use the `container_get_default_ip` helper function to accomplish the same thing in that case:
https://github.com/canonical/microovn/blob/6cf8f907a5505d0d33ddd7bf83343c578866fbbd/tests/test_helper/common.bash#L19-L27

Other things to consider is that the GitHub Hosted runner has very few resources, typically 2 CPUs, so high level of parallelization is likely to cause issues in there.

mkalcok:
There are still services in microovn that are bound to a specific address. That's why rest of the containers passed.
```
ss -ntlp | grep -i ovsdb-server
LISTEN 0      10     10.75.224.203:6644      0.0.0.0:*    users:(("ovsdb-server",pid=14743,fd=23))
LISTEN 0      10     10.75.224.203:6643      0.0.0.0:*    users:(("ovsdb-server",pid=14741,fd=25))
LISTEN 0      10                 *:6642            *:*    users:(("ovsdb-server",pid=14743,fd=29))
LISTEN 0      10                 *:6641            *:*    users:(("ovsdb-server",pid=14741,fd=29))
```
But you are right, I should use your helper for the tests, no reason to have two functions that do almost the same thing.

re: parallelization:
I strongly suspect that a lot of the time in `openssl -connect` is spent waiting on some network I/O so I don't think the parallel execution hurts here.
```
$ time openssl s_client -CAfile /var/snap/microovn/common/data/pki/cacert.pem -cert /var/snap/microovn/common/data/pki/client-cert.pem -key /var/snap/microovn/common/data/pki/client-privkey.pem -quiet -verify_return_error -connect 10.75.224.203:6644
Can't use SSL_get_servername
depth=1 O = MicroOVN, OU = MicroOVN CA, CN = MicroOVN CA
verify return:1
depth=0 O = MicroOVN, OU = ovnsb, CN = mutual-kingfish
verify return:1
{"id":"echo","method":"echo","params":[]}
real	0m4.012s
user	0m0.010s
sys	0m0.004s
```

fnordahl:
> There are still services in microovn that are bound to a specific address. That's why rest of the containers passed.
> 
> ```
> ss -ntlp | grep -i ovsdb-server
> LISTEN 0      10     10.75.224.203:6644      0.0.0.0:*    users:(("ovsdb-server",pid=14743,fd=23))
> LISTEN 0      10     10.75.224.203:6643      0.0.0.0:*    users:(("ovsdb-server",pid=14741,fd=25))
> LISTEN 0      10                 *:6642            *:*    users:(("ovsdb-server",pid=14743,fd=29))
> LISTEN 0      10                 *:6641            *:*    users:(("ovsdb-server",pid=14741,fd=29))
> ```
> 
> But you are right, I should use your helper for the tests, no reason to have two functions that do almost the same thing.

For the services that listen on specific addresses, would not the source of truth for that be the configuration?

Once we introduce tests for containers with multiple interfaces it will most likely matter, so I guess two functions are needed, but I'd still get the IP for the bound services from config and not from `ss`.

If config is unfeasible for some reason, there is some prior art for gleaning similar information from the logs [0] and also a set of wait functions [1] in the OVS testsuites.

0: https://github.com/openvswitch/ovs/blob/269053bf239dd6208acfb9538e4c9713e5b59fc5/tests/ofproto-macros.at#L40
1: https://github.com/openvswitch/ovs/blob/269053bf239dd6208acfb9538e4c9713e5b59fc5/tests/ovs-macros.at#L226



> re: parallelization: I strongly suspect that a lot of the time in `openssl -connect` is spent waiting on some network I/O so I don't think the parallel execution hurts here.

The client and server are on the same host, in the same container, so in reality there is no "network I/O" here, so the fact that this takes so long seems wrong to me. Would be worth investigating if this is a bug either in client, the server or how microovn deploy things?

mkalcok:
>> re: parallelization: I strongly suspect that a lot of the time in openssl -connect is spent waiting on some network I/O so I don't think the parallel execution hurts here.

>The client and server are on the same host, in the same container, so in reality there is no "network I/O" here, so the fact that this takes so long seems wrong to me. Would be worth investigating if this is a bug either in client, the server, the testbed/configuration or how microovn itself deploy things?

You were right. `openssl s_client connect` does not automatically terminate after successfully connecting to the target. The delay we experienced was caused by OVSDB server waiting for our client to send data/requests and eventually timing out and closing the connection. There's a documented way of ending the connection (see [CONNECTED COMMANDS section](https://www.openssl.org/docs/man1.1.1/man1/openssl-s_client.html)) by sending `Q` character to the `openssl` when in non-quiet mode.
```
openssl s_client ... -connect 10.0.0.1:6644 <<< Q
```
Do you want to get rid of the parallelization functions altogether?

> For the services that listen on specific addresses, would not the source of truth for that be the configuration?

While OVN itself is configured by CLI arguments that might not be that easy to parse, maybe we could use `$SNAP_COMMON/data/ovn.env` file that contains `OVN_LOCAL_IP` that is used to generate those CLI args.

fnordahl:
> > > re: parallelization: I strongly suspect that a lot of the time in openssl -connect is spent waiting on some network I/O so I don't think the parallel execution hurts here.
> 
> > The client and server are on the same host, in the same container, so in reality there is no "network I/O" here, so the fact that this takes so long seems wrong to me. Would be worth investigating if this is a bug either in client, the server, the testbed/configuration or how microovn itself deploy things?
> 
> You were right. `openssl s_client connect` does not automatically terminate after successfully connecting to the target. The delay we experienced was caused by OVSDB server waiting for our client to send data/requests and eventually timing out and closing the connection. There's a documented way of ending the connection (see [CONNECTED COMMANDS section](https://www.openssl.org/docs/man1.1.1/man1/openssl-s_client.html)) by sending `Q` character to the `openssl` when in non-quiet mode.
> 
> ```
> openssl s_client ... -connect 10.0.0.1:6644 <<< Q
> ```

Excellent, that looks like a great solution!

> Do you want to get rid of the parallelization functions altogether?

With the solution above, it does not look like we have an immediate need for them for these tests? If you agree I'd rather avid the complexity for now. There is no doubt it was useful to develop them, so time well spent, but let's keep them on the shelf until we find an actually need for them?

> > For the services that listen on specific addresses, would not the source of truth for that be the configuration?
> 
> While OVN itself is configured by CLI arguments that might not be that easy to parse, maybe we could use `$SNAP_COMMON/data/ovn.env` file that contains `OVN_LOCAL_IP` that is used to generate those CLI args.

That does indeed look like a source of the information. You would have to strip off the brackets (`[]`) in case of IPv6 though. There is also the `microovn_get_cluster_address` helper function which gets this information from the output of `microovn status`:
https://github.com/canonical/microovn/blob/f8ae7717bc3c934dbcb3e2340d687169d076e123/tests/test_helper/microovn.bash#L99-L104

mkalcok:
@fnordahl I removed usage of the `parallelize.bash` but I left it in there in case we need it in future. As for the rest, I finished up the tests regarding the TLS and I think it's overall ready for review.

tests/test_helper/parallelize.bash
```diff
@@ -0,0 +1,69 @@
...
+#     # Start multiple commands or functions on the background
+#     start_bg_task "$process_pool" function_1 argument_1 argument_2
+#     start_bg_task "$process_pool" command_2 "$ARG_FROM_VARIABLE"
+#
+#     # Wait for all tasks in the process pool to finish
+#     gather_bg_tasks "$process_pool"
+#     # If any of the tasks failed, this function will return with return code 1
+# }
+
+declare -g -A _PROCESS_POOL
```
fnordahl, line 21: The global variable ought to be prefixed with the name of this file or something similar to minimize chance for collisions.

tests/test_helper/parallelize.bash
```diff
@@ -0,0 +1,69 @@
...
+
+    _PROCESS_POOL["$pool_id"]+="$! "
+}
+
+# gather_bg_tasks PROCESS_POOL
+#
+# Wait for all processes from the PROCESS_POOL to finish. Return 1 if any of the processes
+# had non-zero RC
+function gather_bg_tasks() {
+    local pool_id=$1
```
fnordahl, line 56: Add `; shift` for consistency?

tests/test_helper/microovn.bash
```diff
@@ -44,6 +44,23 @@ function bootstrap_cluster() {
         echo "# Joining cluster with $container" >&3
         lxc_exec "$container" "microovn cluster join $token"
     done
+
+    # Make sure that services are up after the initial bootstrap
```
fnordahl, line 48: The code base contains multiple ways of bootstrapping the cluster, does this perhaps belong in a separate function so that it can be used elsewhere as well?

tests/test_helper/microovn.bash
```diff
@@ -102,3 +119,31 @@ function microovn_get_cluster_address() {
...
+
+    while ! lxc_exec "$container" "lsof -i:$port -sTCP:LISTEN"; do
+        echo "# ($container) waiting for port $port to be opened ($attempt/$max_retry)"
+        if [ $attempt -gt "$max_retry" ]; then
+            echo "# ($container) Maximum retries reached"
+            success=1
+            break
+        fi
+        ((++attempt))
+        sleep 1
```
fnordahl, line 145: One second is a long time, could we use a shorter interval, at least for the first iteration?

tests/test_helper/parallelize.bash
```diff
@@ -0,0 +1,69 @@
...
+#     gather_bg_tasks "$process_pool"
+#     # If any of the tasks failed, this function will return with return code 1
+# }
+
+declare -g -A _PROCESS_POOL
+
+# new_process_pool
+#
+# Create new process pool and print its ID.
+function new_process_pool() {
```
fnordahl, line 26: Bash functions share one global namespace, I'd prefer if we prefix function names with module name to make it apparent where they are defined and what they act on, at least for specialized functions like this.

tests/test_helper/setup_teardown/tls_cluster.bash
```diff
@@ -0,0 +1,31 @@
...
+    load test_helper/microovn.bash
+
+
+    TEST_CONTAINERS=$(container_names "$BATS_TEST_FILENAME" 4)
+    CENTRAL_CONTAINERS=""
+    CHASSIS_CONTAINERS=""
+    counter=0
+    # Categorize first three containers as "CENTRAL" and reset as "CHASSIS"
+    for container in $TEST_CONTAINERS; do
+        if [ $counter -lt 3 ]; then
```
fnordahl, line 13: Assuming the first three containers will always have the central services may not hold true.

Why not just check what services are in each container when running the tests?

tests/test_helper/tls.bash
```diff
@@ -0,0 +1,129 @@
...
+     [$CONTROLLER_CERT_PATH]=$CONTROLLER_KEY_PATH\
+ )
+
+
+# get_ovn_listening_addr CONTAINER
+#
+# Print an IP address on which OVN services are listening in the specified CONTAINER
+function get_ovn_listening_addr() {
+    local container=$1
+    lxc_exec "$container" "source /var/snap/microovn/common/data/ovn.env && echo \$OVN_LOCAL_IP"
```
fnordahl, line 35: This will provide you with an address enclosed in brackets (`[]`) when the system is configured for IPv6.

Why not use something like `microovn_get_cluster_address` instead?

tests/test_helper/tls.bash
```diff
@@ -0,0 +1,129 @@
...
+
+# verify_service_cert CONTAINER IP_ADDR PORT
+#
+# Ensure that service listening on IP_ADDR and PORT uses TLS and can be verified using
+# CA certificate used by the MicroOVN on the specified CONTAINER.
+function verify_service_cert() {
+    local container=$1
+    local ip_addr=$2
+    local port=$3
+    echo "# ($container) Checking TLS on $ip_addr:$port"
```
fnordahl, line 46: Should this output to fd 3 (.e.g `>&3`)?

tests/test_helper/tls.bash
```diff
@@ -0,0 +1,129 @@
...
+    lxc_exec "$container" "openssl s_client -CAfile $CA_CERT_PATH -cert $CLIENT_CERT_PATH -key $CLIENT_KEY_PATH -verify_return_error -connect $ip_addr:$port <<< Q"
```
fnordahl, line 47: This line is very long, please split with `\`

tests/test_helper/setup_teardown/tls_cluster.bash
```diff
@@ -0,0 +1,31 @@
...
+        else
+            CHASSIS_CONTAINERS+="$container "
+        fi
+        ((++counter))
+    done
+
+    export TEST_CONTAINERS
+    export CENTRAL_CONTAINERS
+    export CHASSIS_CONTAINERS
+    launch_containers jammy "${TEST_CONTAINERS[@]}"
```
fnordahl, line 24: Note that I'm in the process of cleaning up the erroneous use of array references to this variable, so might as well do that here as well, ref:
https://github.com/fnordahl/microovn/commit/f08136b4520c0139ac806aa0dbac9e2897beabbb

tests/test_helper/parallelize.bash
```diff
@@ -0,0 +1,69 @@
+# This file contains set of helper functions to run processes or functions
```
fnordahl, line 1: Having unused code in a repository adds unnecessary maintenance burden, should we perhaps postpone adding it until we have a concrete use for it?

fnordahl (changes requested):
Thank you for your work on this!

s/suit/suite in commit message

Please review comments in-line

tests/test_helper/microovn.bash
```diff
@@ -44,6 +44,23 @@ function bootstrap_cluster() {
...
         lxc_exec "$container" "microovn cluster join $token"
     done
+
+    # Make sure that services are up after the initial bootstrap
+    # I believe this can be remove once we resolve https://bugs.launchpad.net/microovn/+bug/2019800
+    # This check can be performed only on cluster members running "central" services
+    local ovn_ports="6641 6642 6643 6644"
+    local counter=0
+    for container in $containers; do
+        if [ $counter -ge 3 ]; then
```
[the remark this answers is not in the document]
mkalcok: Only the first three containers are "OVN Centrals" regardless of how many are deployed. But you are right, approach with `microovn_get_cluster_services` is more versatile.

tests/test_helper/parallelize.bash
```diff
@@ -0,0 +1,69 @@
...
+# gather_bg_tasks PROCESS_POOL
+#
+# Wait for all processes from the PROCESS_POOL to finish. Return 1 if any of the processes
+# had non-zero RC
+function gather_bg_tasks() {
+    local pool_id=$1
+    local ret_code=0
+
+    # Don't exit if 'wait' returns with non-zero RC
+    set +e
```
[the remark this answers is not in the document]
mkalcok: Returning at first error would not collect all background tasks issues requiring more runs to discover all problems. However as suggested I'll remove this file completely as it's not used.

tests/test_helper/parallelize.bash
mkalcok: Removing unused file

tests/test_helper/parallelize.bash
mkalcok: Removing unused file

tests/test_helper/microovn.bash
mkalcok: I was thinking about doing something like exponential backoff but that could add unnecessary complexity. My reasoning for this simple approach was:

- It's dead simple
- Number of retries **roughly** corresponds to the overall timeout in seconds
- `sleep` is invoked only after first failed attempt so reasonably fast machines won't experience it at all
  - This whole check was added only because GH CI was not fast enough with bringing services UP. I was not able to reproduce the error in my local environment. My machine is no beast and I pass this check on first attempt every time, meaning that the 1s sleep is never executed.

If you still think that the 1s is excessive, I'm open to add more granular solution.

tests/test_helper/setup_teardown/tls_cluster.bash
mkalcok: +1. The helper function for checking container services was not available during the initial implementation.

tests/test_helper/tls.bash
mkalcok: Thanks for pointing this out. That function was not available initially.

tests/test_helper/tls.bash
mkalcok: As I mentioned above, I don't think it's necessary to always print messages like this. If the test passes you probably don't care, and if it fails it will be printed.

tests/test_helper/parallelize.bash
mkalcok: Are you OK with global variables from `tls.bash` having `OVN_` prefix? Or should should it be `TLS_OVN_` instead?

tests/test_helper/parallelize.bash
mkalcok: This comment can be relevant in other functions so I'm just wondering what are the benefits of using `shift` if number of arguments is known. For example this:
```bash
function verify_service_cert() {
    local container=$1
    local ip_addr=$2
    local port=$3
    ...
}
```
Looks better than:
```bash
function verify_service_cert() {
    local container=$1
    shift
    local ip_addr=$1
    shift
    local port=$1
    shift
    ...
}
```

tests/test_helper/microovn.bash
```diff
@@ -102,3 +119,31 @@ function microovn_get_cluster_address() {
...
 }
+
+# wait_for_open_port CONTAINER PORT MAX_RETRY
+#
+# Return after specified PORT is open and listening in the CONTAINER.
+#
+# The port can be bound to any interface and it's a good way to test whether
+# OVN services are up and listening. This function will retry for maximum of
+# MAX_RETRY times, each time backing of for 1 second between attempts.
+function wait_for_open_port() {
```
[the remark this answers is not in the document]
mkalcok: Function moved.

tests/test_helper/microovn.bash
```diff
@@ -102,3 +119,31 @@ function microovn_get_cluster_address() {
...
+            echo "# ($container) Maximum retries reached"
+            success=1
+            break
+        fi
+        ((++attempt))
+        sleep 1
+    done
+
+    return $success
+}
```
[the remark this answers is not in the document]
mkalcok: Final newlines added.

tests/test_helper/microovn.bash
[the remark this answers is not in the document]
mkalcok: Replaced with suggestion

tests/test_helper/microovn.bash
mkalcok: Code block extracted to function `wait_ovn_services` in `microovn.bash`

tests/test_helper/setup_teardown/tls_cluster.bash
mkalcok: Replaced with suggestion

tests/test_helper/tls.bash
mkalcok: Function `get_ovn_listening_addr` removed and usage replaced with `microovn_get_cluster_address`

tests/test_helper/tls.bash
mkalcok: Done

tests/test_helper/setup_teardown/tls_cluster.bash
mkalcok: Done

tests/test_helper/parallelize.bash
mkalcok: `parallelize.bash` file removed

tests/test_helper/parallelize.bash
mkalcok: Oh, right, the shifts are inline. I'll push update to be inline with this style, though I'm still curious about rationale behind it.

fnordahl: Thanks for updating!

As for rationale for using `var=$1; shift`:
* The lines declaring receiving variables are independent of each other, a change in the middle later affects one line and not all subsequent lines of code.
* It is quite common to have some variable receive multiple arguments, and we have to shift all previous arguments in order to use `$@` or `$*`.
* It looks more consistent when every declaration of argument receiving variables of the same type look the same.
* Our upstream projects use this style in their test suites.

fnordahl (approved):
LGTM, thanks!

PR #59, opened 2023-08-03
github: Dynamically generate test matrix

fnordahl:
The microovn tests are formulated in a set of test suites.  Each test suite typically deploys an individual namespaced environments for installing the snap, performing configuration and running tests.

On a powerful computer this can be utilized to have the test runner run test suites, and individual tests within each suite in parallel.

However, the GitHub Hosted runners are quite modest and typically provide 2 cores and 7.5 GB RAM.

In order to parallelize jobs in such an environment, we need to run each individual suite in a separate runner.  This is accomplished by using a test matrix.

Dynamically generate the test matrix based on contents of tests/ directory.  This ensures all testsuites are run, and avoids any manual maintenance on addition or removal of a testsuite.

Build the snap once and share the artifact with the system test jobs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,18 +19,16 @@ jobs:
         run: |
           make check-lint
 
-  system-tests:
-    name: System tests
+  build:
+    name: build
     needs:
       - lint
     runs-on: ubuntu-22.04
     env:
-      SNAPCRAFT_BUILD_ENVIRONMENT: "lxd"
+      MICROOVN_SNAP: microovn.snap
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-        with:
-          submodules: recursive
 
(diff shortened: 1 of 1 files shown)


mkalcok:
Wow, this is some serious GHA magic. Overall +1 from me, but I noticed that we seem to have a check duplication. Each check is executed twice. Once with `(pull_request)` suffix and once with `(push)`.

fnordahl:
> Wow, this is some serious GHA magic. Overall +1 from me, but I noticed that we seem to have a check duplication. Each check is executed twice. Once with `(pull_request)` suffix and once with `(push)`.

That part of the workflow was not changed by this patch and is the default behavior.

The reason it runs twice is because I use a branch hosted under canonical/microovn as source for the pull request, when working on a fork it is also useful to get CI feedback before posting the PR.

mkalcok (approved):

PR #60, opened 2023-08-04
Transfer TLS and downscaling content from README to docs

pmatulis:
Transfer some recently-reviewed content from the README to the actual documentation - converting from MD to RST format in the process. This was the material on TLS and downscaling. They each were put in separate How-tos.

Some scaffolding was needed in order for the first How-tos to be accommodated. This consisted of a `how-to` directory and an `index.rst` file.


docs/how-to/tls.rst
```diff
@@ -0,0 +1,173 @@
...
+Re-issue certificates
+~~~~~~~~~~~~~~~~~~~~~
+
+The :command:`certificates reissue` command is used to interact with OVN
+services on the local host; it does not affect peer cluster members.
+
+.. important::
+
+   Services must be running in order to be affected by the
+   :command:`certificates reissue` command.
```
mkalcok, line 72: This note could use a bit of clarification imo.

```suggestion
   Services must be running in order to be affected by the
   :command:`certificates reissue` command. For example, running
   :command:`certificates reissue ovnnb` on a member that does not run this
   service is expected to fail.
```

mkalcok (changes requested):
Thanks for moving this documentation to more appropriate place Peter. I made one suggestion inline, otherwise it looks good imo.

PR #61, opened 2023-08-04
Add low level clustering tests

fnordahl:

Note that this PR depends on #59

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ check-lint: check-tabs
 		-type f \
 		-not -name \*.yaml \
 		-not -name \*.swp \
-		| xargs shellcheck && echo Success!
+		| xargs shellcheck --severity=warning && echo Success!
 
 check-system: $(MICROOVN_SNAP)
 	echo "Running functional tests";					\
--- a/tests/test_helper/bats/cluster.bats
+++ b/tests/test_helper/bats/cluster.bats
@@ -46,3 +46,62 @@ setup() {
         assert [ "$test_family" = "$addr_family" ]
     done
 }
+
+# _test_db_clustered NBSB
+#
+# Tests that database is clustered and listens to the expected address.  The
+# check is run in all containers that lists `central` as one of its services.
+#
+# The NBSB argument can be set to either `nb` or `sb` to indicate which
+# database to check.
+#
+# This test is implemented as a helper function so that we can make use of
+# the bats matrix/parallelization capabilities and is kept in this file
+# because it performs assertions through the `bats-assert` test_helper
+# libraries.
+function _test_db_clustered() {
+    local nbsb=$1; shift
+
+    local cluster_id_str
+
+    for container in $TEST_CONTAINERS; do
+        local container_services
+        container_services=$(microovn_get_cluster_services "$container")
+        if [[ "$container_services" != *"central"* ]]; then
+            echo "Skip $container, no central services" >&3
+            continue
+        fi
+
+        if [ -z "$cluster_id_str" ]; then
+            local cluster_id
+            cluster_id=$(microovn_ovndb_cluster_id "$container" "$nbsb")
+            local cluster_id_abbrev=
+            cluster_id_abbrev=$(echo $cluster_id |cut -c1-4)
+            cluster_id_str="Cluster ID: $cluster_id_abbrev ($cluster_id)"
+        fi
+
+        echo "Checking DB clustered from ${container}'s point of view" >&3
+        local expected_addr
+        expected_addr=$(print_address \
+            "$(microovn_get_cluster_address $container)")
+        local expected_port
+        [ "$nbsb" == "nb" ] && expected_port=6643 || expected_port=6644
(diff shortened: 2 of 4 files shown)


tests/test_helper/bats/cluster.bats
```diff
@@ -46,3 +46,115 @@ setup() {
...
+        fi
+
+        run lxc_exec \
+            "$container" \
+            "microovn.ovs-vsctl get Open_vSwitch . external_ids:ovn-remote"
+        for addr in "${cluster_addresses[@]}"; do
+            expected_addr=$(test_is_ipv6_test && echo "[" || true)
+            expected_addr+=$addr
+            expected_addr+=$(test_is_ipv6_test && echo "]" || true)
+            assert_output -p "$expected_addr"
```
mkalcok, line 119: I'd be weary of partial matching especially wrt IP addresses. If output contains
```
10.0.0.100
10.0.0.101
```
and we are looking for `10.0.0.10` this assertion would pass

tests/test_helper/bats/cluster.bats
mkalcok, line 1: What would you think about taking the non-test functions from this file and putting them to `test_helper/ovn.bash`?

tests/test_helper/bats/cluster.bats
```diff
@@ -46,3 +46,115 @@ setup() {
...
+    assert_equal "${#cluster_addresses[@]}" 3
+    for container in $TEST_CONTAINERS; do
+        run lxc_exec \
+            "$container" \
+            "grep ^$check_var /var/snap/microovn/common/data/ovn.env"
+        for addr in "${cluster_addresses[@]}"; do
+            expected_addr=$(test_is_ipv6_test && echo "[" || true)
+            expected_addr+=$addr
+            expected_addr+=$(test_is_ipv6_test && echo "]" || true)
+            assert_output -p "ssl:${expected_addr}:${expected_port}"
```
mkalcok, line 149: Similar concern about the partial matching, although the accidental match is probably less likely in this case

mkalcok (changes requested):
I left few in-line comments regarding few concerns. Other than that, what would you think about requiring something like python's docstrings for functions and tests?

tests/test_helper/bats/cluster.bats
fnordahl: That is a good point, I was hoping to evade the need for parsing `external_ids:ovn-remote`, but I guess we have to due to this concern, will do and post an update. Thank you for pointing it out!

tests/test_helper/bats/cluster.bats
fnordahl: The string here has equal format to the `external_ids:ovn-remote`, so if we fix that we can re-use the parsing code here.

tests/test_helper/bats/cluster.bats
fnordahl: If you are referring to `_get_member_cluster_address` then yes, I agree.

I'd rather keep the `_test_db_clustered` and `_test_db_connection_string` functions here in the bats file because:
* They are broken out into helper functions only because it allows making use of the bats matrix / parallelization capabilities, and you can see the test definitions only call that one function with a matrix variable
* They perform assertions through `bats-assert`, which is not really available from test_helper libraries

fnordahl:
> I left few in-line comments regarding few concerns. Other than that, what would you think about requiring something like python's docstrings for functions and tests?

Yes, I like that idea.

mkalcok:
I agree. I had something very similar in mind.

tests/test_helper/bats/cluster.bats
fnordahl: Looking at this again, I still think full parsing here, and the extra looping required to make use of it would not be worth the effort. Making use of a fully qualified search string (i.e. `ssl:$expected_addr:6642`) should be more than good enough to avoid false matches.

tests/test_helper/bats/cluster.bats
fnordahl: As commented above, looking at this again the extra effort/looping to get exact matches here is not worth it, the fully qualified search string should be good enough to avoid false matches.

mkalcok (approved):
LGTM.

Makefile
```diff
@@ -11,7 +11,7 @@ check-lint: check-tabs
 		-type f \
 		-not -name \*.yaml \
 		-not -name \*.swp \
-		| xargs shellcheck && echo Success!
+		| xargs shellcheck --severity=warning && echo Success!
```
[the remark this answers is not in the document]
fnordahl: From the commit message:
```
tests: Raise severity level for shellcheck

The `shellcheck` linter provides info-level messages for widely
used shell idioms (SC2015 for example).

These messages may be useful for future travelers, so instead of
suppressing them individually, raise the severity level to warning
in Makefile and subsequently CI.
```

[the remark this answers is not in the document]
mkalcok: thanks

PR #64, opened 2023-08-09
Functional tests for MicroOVN upgrade.

mkalcok:
This test suite ensures that MicroOVN snap can be ugraded from current SnapStore version to the version build for source.

After the upgrade procedure there are bunch of already existing tests that should be executed but it's currently hard to include them without copy-pasting. I'm marking this PR as WIP until we settle on importing/including mechanism


mkalcok:
Oh, it breaks your test matrix generation, but that shouldn't be too hard to solve.

fnordahl:
> I pushed another commit that implements bats file generation. So far it's doing only for `tls_cluster` and `upgrade` tests, just to showcase how it works. More details on the implementation are in commit message. The way I see it:
> 
> Pros:
> 
>     * It's relatively simple implementation
> 
>     * It allows easy reuse of tests between test suits individually or by including entire file

It's definitively an interesting approach to the composition problem and I'm amazed by the simplicity in the implementation. Getting into the business of code generation can be a slippery slope though, so we must be diligent and ensure we don't put too much energy into a bespoke tool that could be spent on improving bats-core itself.

> Cons:
> 
>     * As much as I wanted to use `source` in the bash files for automatic inclusion of other bash files I could as the resulting tests would not be in order and unfortunately we kinda depend on test order with "upgrade" and "scaling" tests. As a solution I implemented relatively simple `# include: file_name` statement that can be processed easily and functions from included files are appended at the end instead of being mixed with the files from original. As a result, recursive inclusion probably does not work.
> 
>     * It's up to a "parent" test to perform necessary "setup" and "setup_file" steps for the included tests to work.

Other thoughts:
* If there is an error raised somewhere in the code, the line numbers cited by the test runner will be in the generated code. This may make it difficult to locate the source of the problem in the sources. Comparing with other tools that do code generation, the autotest framework actually embed source/line references in the generated code so that it is easy to map an error to source-file:line.
* Relying on test execution order is a bit awkward, having said that it does appear to work at the moment. (bats-core allows you to define `BATS_NO_PARALLELIZE_WITHIN_FILE` from within the bats-file to avoid parallel execution within the file which may help)
* Bats-core does not allow duplicate tests, so including tests multiple times would not work (unless we modify their names, but see comment about the slippery slope above).

I took it for a quick spin to see if it would solve the test-morph-test-repeat problem I have in #65, and it does not immediately fit, mainly because of the test duplication issue. I essentially need to run the morph function and all the tests 4 times.

mkalcok:
I updated the PR using your method of test inclusion @fnordahl . It can be reviewed one commit at a time. Sorry for double force push, one of the commits was not signed :|

tests/test_helper/microovn.bash
```diff
@@ -294,3 +314,83 @@ function wait_microovn_online() {
...
+    local ctl_path=$1; shift
+    local db_name=$1; shift
+    local timeout=$1; shift
+    local rc=1
+    local retries=""
+    retries=$((timeout * 2))
+
+    for (( i = 1; i <= "$retries"; i++ )); do
+        echo "# ($container) Waiting for $db_name to apply all changes ($i/$retries)"
+        run lxc_exec "$container" "microovn.ovn-appctl -t $ctl_path cluster/status $db_name"
```
fnordahl, line 338: nit: there is already a function for retrieving the cluster status here:
https://github.com/canonical/microovn/blob/2a0e6b556a02074774dcc5fc50f835a8f2165b50/tests/test_helper/microovn.bash#L148-L150

tests/test_helper/microovn.bash
```diff
@@ -294,3 +314,83 @@ function wait_microovn_online() {
...
+    local retries=""
+    retries=$((timeout * 2))
+
+    for (( i = 1; i <= "$retries"; i++ )); do
+        local container=""
+        local server_present=0
+        for container in $monitor_containers; do
+            local connection_list=""
+            echo "# ($container) Waiting for $target_server_id to depart cluster ($i/$retries)" >&3
+            run lxc_exec "$container" "microovn.ovn-appctl -t $ctl_path cluster/status $db_name"
```
fnordahl, line 377: There is already a function for retrieving the cluster status here:
https://github.com/canonical/microovn/blob/2a0e6b556a02074774dcc5fc50f835a8f2165b50/tests/test_helper/microovn.bash#L148-L150

Any reason to not use that?

tests/test_helper/upgrade_procedures.bash
```diff
@@ -1,3 +1,87 @@
...
+    local target_proto=$1; shift
+    local monitor_containers=$*
+    local ctl_path="/var/snap/microovn/common/run/ovn/"
+    local db_path="/var/snap/microovn/common/data/central/db/"
+    local local_ip=""
+    local_ip=$(microovn_get_cluster_address "$target_container")
+    local target_server_id=""
+    target_server_id=$(microovn_ovndb_server_id "$target_container" "$db_type")
+
+    if [ "$db_type" == "nb" ]; then
```
fnordahl, line 23: There is a function to do this:
https://github.com/canonical/microovn/blob/2a0e6b556a02074774dcc5fc50f835a8f2165b50/tests/test_helper/microovn.bash#L129-L135

It should probably be exported though, and not have a _ prefix

tests/test_helper/microovn.bash
```diff
@@ -294,3 +314,83 @@ function wait_microovn_online() {
...
+# wait_ovsdb_cluster_changes_applied CONTAINER CONTROL_PATH DB_NAME TIMEOUT
+#
+# Wait until OVN OVSDB cluster member converges with rest of the cluster. This function
+# checks output of ovn-appctl to make sure that field 'Entries not yet applied' reaches 0.
+# It requires CONTROL_PATH which points to database's .ctl file (i.e. path/to/ovnnb_db.ctl)
+# and DB_NAME which should be either "OVN_Northbound" or "OVN_Southbound.
+#
+# TIMEOUT in seconds is roughly obeyed. If conditions are not met before timeout is reached, this
+# functions returns non-zero RC
+function wait_ovsdb_cluster_changes_applied() {
```
fnordahl, line 327: nit: I know that it did not exist when you wrote this, just wanted to highlight that there is now a generic `wait_until` function in `common` that we could consider to migrate to using when we get time for it.

tests/test_helper/bats/upgrade.bats
```diff
@@ -19,5 +19,21 @@ setup() {
...
     install_microovn "$MICROOVN_SNAP_PATH" $TEST_CONTAINERS
     perform_manual_upgrade_steps $TEST_CONTAINERS
-    # TODO: include more cluster and TLS tests once we settle on how to include them without copy-pasting
+    local cluster_test_filename="${ABS_TOP_TEST_DIRNAME=}test_helper/bats/post_upgrade_cluster.bats"
+    local tls_test_filename="${ABS_TOP_TEST_DIRNAME=}test_helper/bats/post_upgrade_tls.bats"
+
+    # Note that the outer bats runner will perform validation on the
+    # number of tests ran based on TAP output, so it is important that
+    # the inner bats runner uses a different format for its output.
+    run bats -F junit $cluster_test_filename
```
fnordahl, line 28: nit: The cluster test is IPv6-aware. Should we run both the IPv4 and IPv6 version of it?

The scaleup test accomplishes this with an extra symlink and this in the code:
https://github.com/canonical/microovn/blob/2a0e6b556a02074774dcc5fc50f835a8f2165b50/tests/scaleup_cluster.bats#L60

fnordahl (approved):
Thank you for the great work on this, I love the extensive testing for what will be a documented procedure for early adopters. Very considerate!

There are a few comments in-line, and in the interest of time I will merge it as-is regardless.

PR #67, opened 2023-08-11
ovn/refresh: Don't restart ovn-controller on cluster change

fnordahl:


fnordahl:
Looking at the test failure:
```
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# starttime microovn-scaleup-cluster-2 ovn-controller:             1691736835
```

We need to both wait for services ready before checking and make the collection of starttime bail on failure rather than capturing crime of `/proc`.

mkalcok (approved):
Nice. One less restart on upscale/downscale. I really love these tests!

PR #68, opened 2023-08-11
snap: Build proper version string for snap

fnordahl:

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -3,7 +3,7 @@ icon: microovn.png
 base: core22
 assumes:
  - snapd2.59
-version: git
+adopt-info: ovn
 grade: stable
 source-code: https://github.com/canonical/microovn.git
 summary: Simple clustered OVN deployment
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -183,6 +183,18 @@ parts:
...
+        craftctl default
+        pkg_version=$(
+            dpkg-deb -f \
+            $CRAFT_PART_SRC/../stage_packages/ovn-host*.deb \
+            Version | sed -rne 's/([0-9.]+)[-+].*$$/\1/p')
+        git_version=$(
+            git -C $CRAFT_PROJECT_DIR describe \
+                --always \
+                --dirty \
+                --abbrev=10 | sed -s 's/^v//')
```
mkalcok, line 196: Purely out of curiosity, what's the purpose of this `sed`?

mkalcok (approved):
Just a though here. With the inclusion of the git hash we gain benefit o unique snap version but we'll never get clean `22.03` string. Alternative to this could be that we don't add the hash and in case we need precise version, lets say for debugging purposes, we request snap revision.

However this is purely cosmetic and I'm fine with it going either way

fnordahl:
> Just a though here. With the inclusion of the git hash we gain benefit o unique snap version but we'll never get clean `22.03` string. Alternative to this could be that we don't add the hash and in case we need precise version, lets say for debugging purposes, we request snap revision.
> 
> However this is purely cosmetic and I'm fine with it going either way

On the back of the discussion in #29 I'd rather keep the format we agreed on there. The crucial point for me to include the snap hash is that the underlying upstream version from Ubuntu can change on a snap rebuild without the snap repository having changed.

snap/snapcraft.yaml
fnordahl: Ah, right, it would be useful if we were tagging this repository, to ensure that the `v` in `v1.2.3` does not get exposed in the final version string, but since we are not tagging I can remove it.

Thank you for highlighting it!

PR #71, opened 2023-08-14
Doc: TLS upgrade procedure update

mkalcok:
Added steps that are required to fully transition OVN cluster from `tcp` to `ssl` connection.

### DIFF
--- a/docs/.wordlist.txt
+++ b/docs/.wordlist.txt
@@ -34,3 +34,7 @@ ovs
 vswitchd
 github
 io
+st
+nd
+rd
+th
--- a/docs/.wordlist.txt
+++ b/docs/.wordlist.txt
@@ -34,7 +34,3 @@ ovs
 vswitchd
 github
 io
-st
-nd
-rd
-th
--- a/docs/how-to/tls.rst
+++ b/docs/how-to/tls.rst
@@ -132,15 +132,22 @@ revision of less than ``111``, and there's no way to automatically convert to
 encrypted communication. The following manual steps are needed to upgrade from
 plaintext to TLS:
 
-* ensure that all MicroOVN snaps in the cluster are upgraded to, at least,
-  revision ``111``
-* run ``microovn certificates regenerate-ca`` on one of the cluster members
-* run ``sudo snap restart microovn.daemon`` on **all** cluster members
+1. ensure that all MicroOVN snaps in the cluster are upgraded to, at least,
+   revision ``111``
+
+2. run ``microovn certificates regenerate-ca`` on one of the cluster members
+
+3. run ``sudo snap restart microovn.daemon`` on **all** cluster members
 
 Once this is done, OVN API services throughout the cluster will start listening
 on TLS-secured ports. However, the process is not complete yet because OVN
 Southbound and Northbound database clusters themselves are not capable of
-automatically switching to TLS communication in existing clusters. Both
+automatically switching to TLS communication in existing clusters.
+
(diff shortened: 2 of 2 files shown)


docs/.wordlist.txt
```diff
@@ -34,3 +34,7 @@ ovs
 vswitchd
 github
 io
+st
+nd
+rd
+th
```
mkalcok, line 40: @pmatulis `spellcheck` complained about my usage of `2nd`, `3rd` and `4th` so I added the suffixes to the wordlist, however I'm afraid that this can overshadow actual typos. Any suggestions on how to do this better?

docs/how-to/tls.rst
```diff
@@ -137,8 +137,104 @@ plaintext to TLS:
...
+because we are doing it gradually, cluster data remains intact.
+
+Let's assume that we have a 3 node cluster. We'll start with switching over
+the ``OVN Northbound`` cluster.
+
+**Preparation**: We will be running commands on multiple nodes throughout this
+process, it is recommended to open a separate shell on each node and keep it
+open with following variables exported:
+
+.. code-block:: none
```
pmatulis, line 156: Blank line here

docs/how-to/tls.rst
```diff
@@ -137,8 +137,104 @@ plaintext to TLS:
...
+    CONTROL_SOCKET=/var/snap/microovn/common/run/ovn/ovnnb_db.ctl
```
pmatulis, line 157: Please align text directly under the **c** in `code-block`.

Here, and everywhere else.

docs/how-to/tls.rst
```diff
@@ -137,8 +137,104 @@ plaintext to TLS:
...
+    DB=OVN_Northbound
+    DB_FILE=/var/snap/microovn/common/data/central/db/ovnnb_db.db
+    PORT=6643
+
+**1st step**: Leave cluster with a single member:
+
+.. code-block:: none
+
+    # node-1
```
pmatulis, line 166: Please do not include instructions like this (# comment). Put them in the body of the text.

Here, and everywhere else.

docs/.wordlist.txt
mkalcok: > This new material is great but we may be providing too much information to the user. I suggest we give less context of why we need to perform the steps. If you feel it to be important enough we can provide more detailed content in a Concepts page.

I understand your concern here but this specific doc is slightly special. It describes a situation that affects only people that installed `microovn` from stable channel 'till this day. Anyone who installed it from `edge` or anyone installing it from (default) stable channel starting today, will have have this (TLS) feature enabled automatically. These manual transition steps are really only necessary for small number of current early adopters, but we wanted to give them instructions on how to move on to current stable channel. That's why I think that moving this to the "Concepts page" might not be a good idea.

The reason I went into details here is that there's a lot of repetition in this process. You need to perform 5 steps three times for two clusters and the parameters for these steps vary slightly (almost) each time. That's why my approach was to write these 5 general steps with enough detail that the user will understand what these steps do and how they should be altered each time they are performed.

pmatulis: True, this is only for users on an older snap revision and so wouldn't be appropriate in a Concepts section.

Alright, let's keep it.

mkalcok:
Thanks for the review and suggestions @pmatulis . I incorporated them in separate commit for easier review.

pmatulis (approved):
Thanks Martin! :pie:

PR #73, opened 2023-08-24
docs: Review README and migrate content to Tutorial section

pmatulis:
Review the project README and use its contents to kick-start two new tutorials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,82 +1,10 @@
 # MicroOVN
 
-MicroOVN is snap-deployed OVN with built-in clustering.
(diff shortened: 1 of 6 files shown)


mkalcok (changes requested):
Thanks for migrating the README content into the tutorials section. However I think that since we are removing this information from the README, which is for a lot of people first point of contact with the project, we should put a link to the documentation in its stead.

mkalcok (approved):
Thanks. LGTM. Feel free to merge once the tests pass

PR #75, opened 2023-09-15
docs: Split doc instructions

pmatulis:
Separate the instructions for enabling doc builds (setup.rst) from instructions that doc contributors need (docs/readme.rst). Now we will have:

```
README.md
docs
├── readme.rst
├── setup.rst
```

When users are directed to the `docs` directory from the main README.md they will see how to interact with the documentation, avoiding the bewilderment of configuring the repo for doc builds.

This will be implemented upstream (canonical/sphinx-docs-starter-pack/issues/96) to make this easier for future travellers.

### DIFF
--- a/docs/readme.rst
+++ b/docs/readme.rst
@@ -4,9 +4,9 @@
 Working with the documentation
 ==============================
 
-**Important:** This page is for documentation contributors. It assumes that an
-admin has set up this repository to work with Read the Docs as described in
-file ``setup.rst``.
+**Note:** This page is for documentation contributors. It assumes that an admin
+has enabled this repository to work with Read the Docs as described in file
+``setup.rst``.
 
 There are make targets defined in the ``Makefile`` that set up your local
 environment and allow you to view the documentation. You may also need to
--- a/docs/readme.rst
+++ b/docs/readme.rst
@@ -7,9 +7,9 @@ Working with the documentation
 **Note:** This page is for documentation contributors. It assumes that this
 repository has been enabled for doc builds as described in file ``setup.rst``.
 
-There are make targets defined in the ``Makefile`` file that set up your local
-environment and allow you to view the documentation. You may also need to
-occasionally change the build configuration as new content is added.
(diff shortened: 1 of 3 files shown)


pmatulis:
@fnordahl @mkalcok Please advise on the CI errors. I already re-ran the jobs.

mkalcok:
It looks like there's an issue installing microovn snap with `core24` base. We need to take a look at this because every PR is going to fail. However I don't think we need to gate this PR on functional tests, since it's just a doc change.

docs/readme.rst
```diff
@@ -1,58 +1,20 @@
...
+==============================
+Working with the documentation
+==============================
 
-See the `Sphinx and Read the Docs <https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/>`_ guide for instructions on how to get started with Sphinx documentation.
+**Note:** This page is for documentation contributors. It assumes that this
+repository has been enabled for doc builds as described in file ``setup.rst``.
 
-Then go through the following sections to use this starter pack to set up your docs repository.
+There are make targets defined in the ``Makefile`` file that set up your local
```
mkalcok, line 10: We have multiple `Makefile`s in this repository, so we should be more specific here. i.e.: `docs/Makefile`

mkalcok (changes requested):
Thank you for the update Peter. I left one small suggestion inline, otherwise it looks good.

docs/readme.rst
pmatulis: I will make it more explicit but I cannot give a path because the doc stuff can apply to a non-code-repository (`Makefile` is in the root) and a code repository may conceivably use a directory other than `docs`.

mkalcok: Do you mean that there's a use case in which documentation sources are distributed without rest of this repository?

pmatulis: So I'm taking a generic approach here. I could have provided a path **for this current repository** but the docs starter pack upon which all this is based can be used within a standalone doc repository where there *is no* docs directory. It's all in the root.

PR #77, opened 2023-09-28
snap/wrappers: Be selective on libraries included for `jq`.

fnordahl:
The `jq` stage-package was added in commit 6262a1d0e278, and it pulls in several dependencies.  Trim down to what is actually needed.


mkalcok (approved):
LGTM, thanks for this optimization. interesting that that are unneeded dependencies in `jq`

PR #83, opened 2023-10-19
Split `microovn.central` service.

mkalcok:
Previously `microovn.central` started all three central services (ovn-nb, ovn-sb and northd). Now these daemons are split into their own services for more fine-grained controll over them.

The `microovn.central` will stay for now and act as an umbrella service that can trigger start of all three services. However the usage is discouraged as it will be removed. It's also not capable of being used as shortcut for stopping these services.

Additional consequence of splitting of the services is that we can now scale the cluster without restarting OVN NB/SB services as only the northd daemon needs restart.

### DIFF
--- a/docs/how-to/tls.rst
+++ b/docs/how-to/tls.rst
@@ -112,6 +112,8 @@ member.
    member. Any failure will result in subsequent communication errors for that
    service within the cluster.
 
+.. _certificates_lifecycle:
+
 Certificate lifecycle
 ---------------------
 
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -1,0 +1,11 @@
+=========
+Reference
+=========
+
+MicroOVN reference material is specific to the MicroOVN project. It
+does not cover upstream OVN/OVS topics.
+
+.. toctree::
+   :maxdepth: 1
+
+   services
--- a/docs/reference/services.rst
+++ b/docs/reference/services.rst
@@ -1,0 +1,71 @@
+=================
+MicroOVN services
+=================
+
+This page presents a list of all MicroOVN services. Their descriptions are
+for reference only - the user is not expected to interact directly with these
+services.
+
+The status of all services is displayed by running:
+
+.. code-block:: none
+
+   snap services microovn
+
+``microovn.central``
+--------------------
+
+.. warning::
+
+   The ``microovn.central`` service is deprecated and will be removed in a
+   future release.
+
+This is a transitional service. Starting this service will start and enable
+multiple services:
+
+* ``microovn.ovn-ovsdb-server-nb``
+* ``microovn.ovn-ovsdb-server-sb``
+* ``microovn.ovn-northd``
+
+However this service is not capable of stopping these child services so its
+usage is strongly discouraged. Users should use individual services instead.
+
+``microovn.chassis``
+--------------------
+
+This service maps directly to the ``ovn-controller`` daemon.
+
+``microovn.daemon``
+-------------------
+
+The main MicroOVN service/process that manages all the other processes. It also
+handles communication with other MicroOVN cluster members and provides an API
+for the ``microovn`` client command.
+
+``microovn.ovn-ovsdb-server-nb``
+--------------------------------
+
+This service maps directly to the ``OVN Northbound`` database/service.
+
+``microovn.ovn-northd``
+-----------------------
+
+This service maps directly to the ``ovn-northd`` daemon.
+
+``microovn.ovn-ovsdb-server-sb``
+--------------------------------
+
+This service maps directly to the ``OVN Southbound`` database/service.
+
+``microovn.refresh-expiring-certs``
+-----------------------------------
+
+This service is a recurring process that runs once a day between ``02:00`` and
+``02:30``. It triggers TLS certification reissue for certificates that are
+nearing the expiration. For more information see the
+:ref:`certificates lifecycle <certificates_lifecycle>`.
+
+``microovn.switch``
+-------------------
+
+This services maps directly to the ``ovs-vswitchd`` daemon.
--- a/microovn/ovn/bootstrap.go
+++ b/microovn/ovn/bootstrap.go
@@ -93,9 +93,19 @@ func Bootstrap(s *state.State) error {
 	}
 
 	// Enable OVN central.
-	err = snapStart("central", true)
+	err = snapStart("ovn-ovsdb-server-nb", true)
 	if err != nil {
-		return fmt.Errorf("Failed to start OVN central: %w", err)
+		return fmt.Errorf("Failed to start OVN NB: %w", err)
+	}
+
+	err = snapStart("ovn-ovsdb-server-sb", true)
+	if err != nil {
+		return fmt.Errorf("Failed to start OVN SB: %w", err)
+	}
+
+	err = snapStart("ovn-northd", true)
+	if err != nil {
+		return fmt.Errorf("Failed to start OVN northd: %w", err)
 	}
 
 	// Generate certificate for OVN chassis (controller)
--- a/microovn/ovn/bootstrap.go
+++ b/microovn/ovn/bootstrap.go
@@ -68,9 +68,9 @@ func Bootstrap(s *state.State) error {
 	// all OVS-based programs ability to specify active or passive (listen)
 	// connection types.
 	err = GenerateNewServiceCertificate(s, "client", CertificateTypeServer)
-        if err != nil {
-                return fmt.Errorf("failed to generate TLS certificate for client: %s", err)
-        }
+	if err != nil {
+		return fmt.Errorf("failed to generate TLS certificate for client: %s", err)
+	}
 
 	// Enable OVS switch.
 	err = snapStart("switch", true)
--- a/snapcraft/commands/central.start
+++ b/snapcraft/commands/central.start
@@ -1,58 +1,15 @@
 #!/bin/sh
-set -eux
 
-# Load the environment
-. "${SNAP_COMMON}/data/ovn.env"
+echo "This is a transitional service that can be used to start 'ovn-ovsdb-server-nb',
+'ovn-ovsdb-server-sb' and 'ovn-northd' services at the same time. However its usage is
+discouraged as it will be removed in future releases."
 
-# Setup directories
-export OVN_DBDIR="${SNAP_COMMON}/data/central/db"
-export OVN_LOGDIR="${SNAP_COMMON}/logs"
-export OVN_RUNDIR="${SNAP_COMMON}/run/ovn"
-export OVN_PKGDATADIR="${SNAP}/share/ovn"
-export OVN_SYSCONFDIR="${SNAP}/etc"
-export OVN_PKIDIR="${SNAP_COMMON}/data/pki"
-
-# Prepare the arguments
-# By specifying "--db-*-create-insecure-remote=no" we prevent creation of
-# hardcoded bindings and we can use database to configure remotes later.
-OVN_ARGS="--db-nb-addr="${OVN_LOCAL_IP}" \
---db-nb-create-insecure-remote=no \
---db-sb-addr="${OVN_LOCAL_IP}" \
---db-sb-create-insecure-remote=no \
---db-nb-cluster-local-addr="${OVN_LOCAL_IP}" \
---db-sb-cluster-local-addr="${OVN_LOCAL_IP}" \
---ovn-northd-nb-db="${OVN_NB_CONNECT}" \
---ovn-northd-sb-db="${OVN_SB_CONNECT}" \
---db-nb-cluster-local-proto=ssl \
---db-nb-cluster-remote-proto=ssl \
---db-sb-cluster-local-proto=ssl \
---db-sb-cluster-remote-proto=ssl \
---ovn-northd-ssl-key="${OVN_PKIDIR}"/ovn-northd-privkey.pem \
---ovn-northd-ssl-cert="${OVN_PKIDIR}"/ovn-northd-cert.pem \
---ovn-northd-ssl-ca-cert="${OVN_PKIDIR}"/cacert.pem \
---ovn-nb-db-ssl-key="${OVN_PKIDIR}"/ovnnb-privkey.pem \
---ovn-nb-db-ssl-cert="${OVN_PKIDIR}"/ovnnb-cert.pem \
---ovn-nb-db-ssl-ca-cert="${OVN_PKIDIR}"/cacert.pem \
---ovn-sb-db-ssl-key="${OVN_PKIDIR}"/ovnsb-privkey.pem \
---ovn-sb-db-ssl-cert="${OVN_PKIDIR}"/ovnsb-cert.pem \
---ovn-sb-db-ssl-ca-cert="${OVN_PKIDIR}"/cacert.pem"
-
-if [ "${OVN_INITIAL_NB}" != "${OVN_LOCAL_IP}" ]; then
-    OVN_ARGS="${OVN_ARGS} --db-nb-cluster-remote-addr="${OVN_INITIAL_NB}""
-fi
-
-if [ "${OVN_INITIAL_SB}" != "${OVN_LOCAL_IP}" ]; then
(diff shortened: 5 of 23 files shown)


snapcraft/ovn-central.env
```diff
@@ -0,0 +1,21 @@
...
+export OVN_SYSCONFDIR="${SNAP}/etc"
+export OVN_PKIDIR="${SNAP_COMMON}/data/pki"
+
+# Disable some commands
+mkdir -p "${OVN_RUNDIR}/bin/"
+for i in install plymouth sudo systemctl; do
+    [ -e "${OVN_RUNDIR}/bin/${i}" ] && continue
+
+    ln -s "/bin/true" "${OVN_RUNDIR}/bin/${i}"
+done
```
fnordahl, line 20: While I understand how these ended up in this file, it's probably not the correct place to host them after the split.

Having three services starting independently attempting to create this might lead to a race conditions and spurious errors. (And having an sourceable "environment" file performing actions behind your back comes as an element of surprise.)

Would it make sense to move this to either the install hook or as part of the startup of the `daemon` service?

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+    snap services microovn
+
+microovn.central
+----------------
+
+.. warning::
+
+    Deprecated: This service will be removed in future release
+
+This is a shortcut service. Starting this service will start and enable
```
fnordahl, line 20: s/shortcut/transitional

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+``microovn.ovn-nb``, ``microovn.ovn-sb`` and ``microovn.ovn-northd``. However
+this service , due to the ``snap`` limitations, is not capable of stopping
+child services and its usage is generally discouraged.
```
fnordahl, line 23: I'd rather highlight the reasons for why managing these services separately is desirable rather than bashing snaps here

snapcraft/commands/central.start
```diff
@@ -1,67 +1,15 @@
...
-
-# Start SouthBound OVN DB
-"${SNAP}/share/ovn/scripts/ovn-ctl" run_sb_ovsdb ${OVN_ARGS} &
-
-# Start NorthBOund daemon
-"${SNAP}/share/ovn/scripts/ovn-ctl" start_northd ${OVN_ARGS} \
-    --ovn-manage-ovsdb=no --no-monitor
-
-sleep infinity
+echo "This service can be used as a shortcut to start ovn-nb, ovn-sb and \
```
fnordahl, line 3: s/shortcut/transitional

tomponline (commented):
I'm looking forward to this landing :)

microovn/ovn/bootstrap.go
```diff
@@ -68,9 +68,9 @@ func Bootstrap(s *state.State) error {
 	// all OVS-based programs ability to specify active or passive (listen)
 	// connection types.
 	err = GenerateNewServiceCertificate(s, "client", CertificateTypeServer)
-        if err != nil {
-                return fmt.Errorf("failed to generate TLS certificate for client: %s", err)
-        }
+	if err != nil {
+		return fmt.Errorf("failed to generate TLS certificate for client: %s", err)
+	}
```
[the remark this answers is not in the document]
mkalcok: MicroOVN seems to be using tabs for indentation and this piece of code used spaces.This was an auto-formater change and I'd leave it in for consistency.

microovn/ovn/bootstrap.go
```diff
@@ -93,9 +93,19 @@ func Bootstrap(s *state.State) error {
 	}
 
 	// Enable OVN central.
-	err = snapStart("central", true)
+	err = snapStart("ovn-nb", true)
```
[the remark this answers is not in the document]
mkalcok: Ideally users should not interact with these services at all...but I see your point and I don't see any downside to this change, aside from bit of verbosity

microovn/ovn/bootstrap.go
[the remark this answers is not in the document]
fnordahl: Cleaning up inconsistencies is always a nice thing, but perhaps they belong in a separate commit?

snapcraft/ovn-central.env
fnordahl: I'd rather not have them in a sourceable `.env` file.

Speaking of who needs this, I honestly do not think this is related to any of the central services at all, I do not find any evidence of the `ovn-ctl`, `ovs-ctl` or `ovs-lib` scripts making calls to these binaries. So I wonder what was the reason for the original introduction of this, habits of the original author?

docs/reference/services.rst
fnordahl: Sure, I'd angle it with a message telling the user to use the other services to stop things instead then. Together with the transitional label it will make it absolutely clear to the user that they should stop whatever it is they think they should be doing and follow the new workflow

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
+=================
+MicroOVN Services
```
pmatulis, line 2: s/Services/services

docs/reference/index.rst
```diff
@@ -0,0 +1,11 @@
+==========
+References
+==========
+
+These references describe concepts unique to MicroOVN. They do not aim to
+describe general OVN/OVS concepts
+
```
pmatulis, line 7: MicroOVN reference material is specific to the MicroOVN project. It does not cover upstream OVN/OVS topics.

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+=================
+
+This document provides brief explanation of services provided by MicroOVN. They
+are generally not meant for direct user interaction and should be used only
+if necessary. List and status of these services can be displayed by running:
+
```
pmatulis, line 8: This page presents a list of all MicroOVN services. The descriptions presented here is for reference only - the user is not expected to interact directly with these services.

The status of all services is displayed by running:

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+.. code-block:: none
+
+    snap services microovn
```
pmatulis, line 11: Align the first letter of the command ('s') directly under the 'c' (of code-block). This should be done everywhere in the document, even for 'note' or 'warning', etc.

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+microovn.central
+----------------
+
+.. warning::
+
+    Deprecated: This service will be removed in future release
```
pmatulis, line 18: The ``microovn.central`` service is deprecated and will be removed in a future release.

docs/reference/services.rst
pmatulis, line 20: ... and enable multiple services:

* ``microovn.ovn-nb``
* ``microovn.ovn-sb``
* ``microovn.ovn-northd``

However, ...

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+This document provides brief explanation of services provided by MicroOVN. They
+are generally not meant for direct user interaction and should be used only
+if necessary. List and status of these services can be displayed by running:
+
+.. code-block:: none
+
+    snap services microovn
+
+microovn.central
```
pmatulis, line 13: Let's put this in monospace (double backticks) - same for the rest of the headers with service names.

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+This is a shortcut service. Starting this service will start and enable
+``microovn.ovn-nb``, ``microovn.ovn-sb`` and ``microovn.ovn-northd``. However
+this service , due to the ``snap`` limitations, is not capable of stopping
+child services and its usage is generally discouraged.
+
+microovn.chassis
+----------------
+
+This service maps directly to ``ovn-controller`` daemon.
```
pmatulis, line 28: ... the ``ovn-controller``

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+microovn.daemon
+---------------
+
+Main MicroOVN service/process that manages rest of the processes. It also
```
pmatulis, line 33: The main ...

... manages all the other processes.

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+handles communication with other MicroOVN cluster members and provides an API
+for ``microovn`` client command.
```
pmatulis, line 35: ... for the ``microovn`` ...

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+microovn.ovn-nb
+---------------
+
+This service maps directly to ``OVN Northbound`` database/service.
```
pmatulis, line 40: ... directly to the ...

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+microovn.ovn-northd
+-------------------
+
+This service maps directly to ``ovn-northd`` daemon.
```
pmatulis, line 45: ... directly to the ...

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+microovn.ovn-sb
+---------------
+
+This service maps directly to ``OVN Southbound`` database/service.
```
pmatulis, line 50: ... directly to the ...

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+
+microovn.refresh-expiring-certs
+-------------------------------
+
+This service is a recurring process that runs once a day between ``02:00`` and
+``02:30``. It triggers TLS certification reissue for certificates that are
+nearing the expiration. For more information see ``Certificate lifecycle``
```
pmatulis, line 57: ... information see the ...

docs/reference/services.rst
```diff
@@ -0,0 +1,63 @@
...
+section in :doc:`/how-to/tls`
```
pmatulis, line 58: Link directly to that section. See:

https://docs.openstack.org/charm-guide/latest/community/doc-contrib/rst.html#linking-to-a-location-within-a-page-in-the-doc-set

pmatulis (changes requested):
Thanks for this. See change requests.

I would normally expect to see a significant doc change like this as a separate PR with a separate Jira card.

snapcraft/ovn-central.env
mkalcok: I'd be in favor of removing this then. The symlinks do not even have effect without the very specific `{$OVN_RUNDIR}/bin/` added to path, and that happens only here.

docs/reference/index.rst
```diff
@@ -0,0 +1,11 @@
+==========
+References
```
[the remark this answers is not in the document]
mkalcok: I followed naming convention in other documentation modes. Both `How-to guides` and `Tutorials` use plural.

mkalcok:
Thank you for extensive review @fnordahl @pmatulis . I believe I addressed all the comments.

> I would normally expect to see a significant doc change like this as a separate PR

The need for this documentation was created directly by code changes in this PR. We split the `microovn.central` service but retained it (for some underlying technical reasons) which, imo, created a bit of an ambiguity. That's why we decided to include also this documentation piece that would explain the purpose of each service exposed by the MicroOVN snap.

docs/reference/services.rst
```diff
@@ -0,0 +1,71 @@
...
+
+   The ``microovn.central`` service is deprecated and will be removed in a
+   future release.
+
+This is a transitional service. Starting this service will start and enable
+multiple services:
+
+* ``microovn.ovn-ovsdb-server-nb``
+* ``microovn.ovn-ovsdb-server-sb``
+* ``microovn.ovn-northd``.
```
pmatulis, line 28: extra dot

docs/reference/services.rst
```diff
@@ -0,0 +1,71 @@
+=================
+MicroOVN services
+=================
+
+This page presents a list of all MicroOVN services. The descriptions presented
```
pmatulis, line 5: apologies, my fault here:

s/The descriptions presented here is/Their descriptions are/

pmatulis (changes requested):
Sorry, a few last things

snapcraft/commands/central.start
```diff
@@ -1,67 +1,15 @@
...
-export OVN_PKGDATADIR="${SNAP}/share/ovn"
-export OVN_SYSCONFDIR="${SNAP}/etc"
-export OVN_PKIDIR="${SNAP_COMMON}/data/pki"
-
-# Disable some commands
-mkdir -p "${OVN_RUNDIR}/bin/"
-for i in install plymouth sudo systemctl; do
-    [ -e "${OVN_RUNDIR}/bin/${i}" ] && continue
-
-    ln -s "/bin/true" "${OVN_RUNDIR}/bin/${i}"
```
fnordahl, line 20: Good riddance!

However, I see these are also in `snapcraft/commands/chassis.start` and `snapcraft/commands/switch.start`.

Would it be more consistent to drop all of these simultaneously in a separate parent commit?


For the record, there are now some additional `DENIED` messages logged, but they do not affect operation.
```
[  101.595042] audit: type=1400 audit(1697770466.057:241): apparmor="DENIED" operation="exec" profile="snap.microovn.ovn-ovsdb-server-nb" name="/usr/bin/systemctl" pid=2804 comm="ovn-ctl" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
[  101.595086] audit: type=1400 audit(1697770466.057:242): apparmor="DENIED" operation="exec" profile="snap.microovn.ovn-ovsdb-server-nb" name="/usr/bin/systemctl" pid=2804 comm="ovn-ctl" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
[  101.596966] audit: type=1400 audit(1697770466.061:243): apparmor="DENIED" operation="exec" profile="snap.microovn.ovn-ovsdb-server-nb" name="/usr/bin/plymouth" pid=2806 comm="ovn-ctl" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
[  101.597023] audit: type=1400 audit(1697770466.061:244): apparmor="DENIED" operation="exec" profile="snap.microovn.ovn-ovsdb-server-nb" name="/usr/bin/plymouth" pid=2806 comm="ovn-ctl" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
```

Even without this change there are still other unhandled `DENIED` rules logged, such as:
```
[  101.599507] audit: type=1400 audit(1697770466.065:245): apparmor="DENIED" operation="capable" profile="snap.microovn.switch" pid=2787 comm="ovsdb-server" capability=38  capname="perfmon"
```

So I still think we should still drop these hacks and figure out what indirect dependencies of the OVS/OVN ctl scripts end up calling the offending binaries and provide patches upstream at some point in the future.

docs/how-to/tls.rst
```diff
@@ -112,6 +112,8 @@ member.
    member. Any failure will result in subsequent communication errors for that
    service within the cluster.
 
+.. _certificates_lifecycle:
```
fnordahl, line 115: Move the doc changes to a separate commit? Some of these changes will surely cause conflicts on backport, and it would be very nice to not have to deal with conflicts on the actual code changes.

fnordahl (changes requested):
I took this for a spin and tested both refresh, fresh install and post configuration reboots, looks great.

The commit messages may need updating to reflect changes during review.

I also have a couple of comments in-line that request splitting a couple of things into separate commits.

snapcraft/commands/central.start
mkalcok: I agree. I'll drop those instances too. The commands do not work anyway, it's better to have the failures logged rather than silently failing.

docs/how-to/tls.rst
mkalcok: Yeah. I didn't slice this PR into commits very well.

mkalcok:
Thanks again for the review. I incorporated your comments and I separated the docs and the command masking removal into separate commits.

fnordahl (approved):
LGTM, thanks!

PR #88, opened 2023-10-24
docs: Update build and configuration files

pmatulis:
Update the build and configuration files for documentation. This is based on the latest from the sphinx-docs-starter-pack repository. Notable changes include:

* updates to docs `readme` (and the first sub-page `help-woke` to keep things tidy; there will be more in the future)
* `custom_conf.py` (custom project main configuration file)
* `.custom_wordlist.txt` (custom project false positive word spelling errors)
* page header customisation
* project icon

The header contains more information now:

image

The file `microovn.png` in the project root was used in the `docs` directory (copied it):

image


docs/.custom_wordlist.txt
```diff
@@ -0,0 +1,26 @@
...
+favicon
+installable
+Makefile
+yaml
+RTD
+readthedocs
+MicroOVN
+ovsdb
+OVN
+northd
```
fnordahl, line 15: Just a comment, no action needed: I'm a bit surprised that spelling passes with this revised list (i did check locally), given it is much shorter than the original, but I guess the updates changes settings so that some of the items in the word list are no longer necessary.

fnordahl (commented):
Thanks for working on this, Peter!

LGTM, however there is one administrative change that is required in order for us to merge this.

In preparation for making use of self-hosted GitHub Actions runners, we recently enabled a branch protection rule that [require signed commits](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-signed-commits).

If you already have GPG set up it should be as easy as `git commit -S`, to fix up this branch you can do a interactive rebase and amend the individual commits with a signature.

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

docs/.custom_wordlist.txt
mkalcok: There are two wordlists now, `.wordlist.txt` and `custom_wordlist.txt` and together they are larger than previous `.wordlist.txt`. My guess is that the `wordlist` is supposed to be a direct copy from documentation starterpack repo and `custom_wordlist` is for project-specific words.

fnordahl: That is my assumption too, but the combination of the two lists still miss a lot of project specific words in the original `.wordlist.txt`, with still passing checks, so something else has changed simultaneously.

mkalcok (approved):
LGTM. Thanks for the updates.

PR #91, opened 2023-11-02
Cleanup OVS resource on snap removal.

mkalcok:
This change modifies `microovn.switch` service to stop OVS vswitchd service with `--cleanup` flag. This ensures that when snap is removed, datapath resources like ports and bridges are cleaned up.

This new behavior is overriden in pre-refresh hook to avoid dataplane outages during the snap refresh.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -75,7 +75,7 @@ apps:
     command: commands/switch.start
     daemon: simple
     install-mode: disable
-    refresh-mode: endure
+    stop-command: commands/ovs-appctl exit --cleanup
     plugs:
       - firewall-control
       - hardware-observe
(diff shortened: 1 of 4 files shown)


snap/snapcraft.yaml
```diff
@@ -76,6 +76,7 @@ apps:
     daemon: simple
     install-mode: disable
     refresh-mode: endure
```
fnordahl, line 78: Now that we handle stop/refresh carefully we can remove `refresh-mode` 'endure'.

tests/removal.bats
```diff
@@ -0,0 +1 @@
+test_helper/bats/removal.bats
```
fnordahl, line 1: I wonder if we should rename the new test files and use a more generic term. For example lifecycle?

We could then at some point in the future extend the testing to cover the expected handling to `snap stop`, `snap refresh` in addition to `remove`.

WDYT?

fnordahl (changes requested):
Thank you for working on this!

The approach generally looks good to me, I only have a comment on naming of the new test modules.

I also would like to see similar handler for the `chassis` service, but that can be done as a follow-up.

fnordahl (approved):
LGTM, thanks!

PR #93, opened 2023-11-03
snap: Handle stop and refresh for chassis service.

fnordahl:


mkalcok (approved):
LGTM. Thank you for figuring out the issue with cleanly stopping the switch.

I'm partly concerned about each wrapper setting it's own env vars, but that's a broader issue across the wrappers so I opened [LP 2042843](https://bugs.launchpad.net/microovn/+bug/2042843)

PR #94, opened 2023-11-04
Use systemd journal for logging.

fnordahl:
At present the OVS and OVN services provided by MicroOVN logs to file.

This is a problem, because there is no mechanism for rotating these logfiles, and as a consequence they will grow without bounds potentially depleting all available storage.

The best practice for snaps is to log to the systemd journal [0].

Due to how the `ovs-ctl` and `ovn-ctl` scripts work, there is no way to avoid the log files from being created.  It may also be useful to have log files configured for live debugging sessions.

Add a README in $SNAP_COMMON/logs explaining that the log files are not updated by default, pointing the user to where to find the actual logs.

0: https://forum.snapcraft.io/t/best-practice-for-logging/1210

Closes-Bug: #2039582

### DIFF
--- a/docs/how-to/logs.rst
+++ b/docs/how-to/logs.rst
@@ -1,0 +1,56 @@
+==============
+Accessing logs
+==============
+
+The :ref:`MicroOVN services` provide logs as part of their normal operation.
+
+By default they are provided through the systemd journal.
+
+Below you can view an example for how to access the logs of the
+``microovn.chassis`` service:
+
+.. code-block:: none
+
+   journalctl -u microovn.chassis
+
+Below you can view an example of how to view a live log display for the same
+service:
+
+.. code-block:: none
+
+   journalctl -f -u microovn.chassis
+
+Log files
+---------
+
+Inside the ``/var/snap/microovn/common/logs`` directory you will find files for
+each individual service, however these will either be empty or not contain
+updated information, this is intentional.
+
+On fresh install the files are created, as a precaution, in the event a need
+arises for enabling `debug logging`_.  On refresh from older versions of
+MicroOVN existing files will be retained, but not updated.
+
+Debug logging
+-------------
+
+The Open vSwitch (OVS) and Open Virtual Network (OVN) daemons have a rich set
+of debug features, one of which being available by augmenting log levels for
+individual modules at run time.
+
+A list of modules can be acquired through the ``microovn.ovs-appctl`` and
+``microovn.ovn-appctl`` commands.
+
+Below you can view an example of how to enable debug logging for the Open
+vSwitch ``vswitchd`` module:
+
+.. code-block:: none
+
+   microovn.ovs-appctl vlog/set vswitchd:file:dbg
+
+Below you can view an example of how to enable debug logging for the Open
+Virtual Network ``reconnect`` module:
+
+.. code-block:: none
+
+   microovn.ovn-appctl vlog/set reconnect:file:dbg
--- a/docs/how-to/logs.rst
+++ b/docs/how-to/logs.rst
@@ -6,15 +6,13 @@ The :ref:`MicroOVN services` provide logs as part of their normal operation.
 
 By default they are provided through the systemd journal.
 
-Below you can view an example for how to access the logs of the
-``microovn.chassis`` service:
+This is how you can access the logs of the ``microovn.chassis`` service:
 
 .. code-block:: none
 
    journalctl -u microovn.chassis
 
-Below you can view an example of how to view a live log display for the same
-service:
+This is how you can view a live log display for the same service:
 
 .. code-block:: none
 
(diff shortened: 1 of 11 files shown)


mkalcok (dismissed):
LGTM

pmatulis:
The RTD build should now be done (after making a change on the RTD side). Maybe submitting a trivial change here will "kick" everything in the right direction. I tried re-running the Doc job but the expected build did not happen.

fnordahl:
> The RTD build should now be done (after making a change on the RTD side). Maybe submitting a trivial change here will "kick" everything in the right direction. I tried re-running the Doc job but the expected build did not happen.

Thanks, Peter! Done

docs/how-to/logs.rst
```diff
@@ -0,0 +1,62 @@
...
+-------------
+
+The Open vSwitch (OVS) and Open Virtual Network (OVN) daemons have a rich set
+of debug features, one of which being available by augmenting log levels for
+individual modules at run time.
+
+A list of modules can be acquired through the ``microovn.ovs-appctl`` and
+``microovn.ovn-appctl`` command line utilities, for example:
+
+.. code-block:: none
```
pmatulis, line 44: is the output to these two list commands extensive? generally it is useful to include output to commands if the output is reasonable

pmatulis (changes requested):
Comment on command output

docs/how-to/logs.rst
mkalcok: There's a lot of modules that can be individually configured. It outputs ~120 lines

fnordahl: I agree with @mkalcok the output is quite long/verbose. Would a short/edited version of it be valuable just to get a feel of it, alternatively we could add a reference page with this information and link to that?

fnordahl: (I guess an alternative could also be to omit the code block for those two commands)

pmatulis: Often we do stuff like this:

```
first line output
second line output
...
penultimate line output
last line output
```

mkalcok: @pmatulis The examples of `vlog/get` commands got taken out, the doc just generically points user towards `microovn.ovs-appctl `and `microovn.ovn-appctl` commands, which I think is fine. We don't need to re-document usage of the general OVN/OVS commands.
Examples of how to `vlog/set` values are still in there but they don't produce any output.

Is the current form OK with you?

docs/how-to/logs.rst
```diff
@@ -0,0 +1,56 @@
...
+
+   microovn.ovs-appctl vlog/set vswitchd:file:dbg
+
+Below you can view an example of how to enable debug logging for the Open
+Virtual Network ``reconnect`` module:
+
+.. code-block:: none
+
+   microovn.ovn-appctl vlog/set reconnect:file:dbg
```
pmatulis, line 56: *Is it just debug logging can be enabled and disabled or are there other levels?*

pmatulis (changes requested):

docs/how-to/logs.rst
mkalcok: There are other levels too. The format, I believe, is `<module>:<output_target>:<level>`

docs/how-to/logs.rst
```diff
@@ -0,0 +1,56 @@
+==============
+Accessing logs
+==============
+
+The :ref:`MicroOVN services` provide logs as part of their normal operation.
+
+By default they are provided through the systemd journal.
+
+Below you can view an example for how to access the logs of the
```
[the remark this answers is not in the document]
mkalcok: Just for my curiosity, is the phrase `This is how to ...` more standard across our documentation? I'm just asking because I don't see much difference between the two, `Below you can ...` has even fewer words :laughing:

docs/how-to/logs.rst
```diff
@@ -0,0 +1,56 @@
...
+.. code-block:: none
+
+   journalctl -u microovn.chassis
+
+Below you can view an example of how to view a live log display for the same
+service:
+
+.. code-block:: none
+
+   journalctl -f -u microovn.chassis
```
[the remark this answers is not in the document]
mkalcok: I think that the full example might be justified here. The `-u` flag needs to be followed by the unit name, so for example, `journalctl -u -f microovn.chassis` would not work.
What do you think?

docs/how-to/logs.rst
fnordahl: There is a lot we could document about logging, just want to remind you that the main motivation for creating this page is to get a `README` file that we can put into `/var/snap/microovn/common/logs` to explain why there are empty or not updated log files there which is a consequence of this change.

As explained in the various commit messages we need to do this to solve a real risk of depleting the storage on end users machines, and there is currently no way to not have the files created.

docs/how-to/logs.rst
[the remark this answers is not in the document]
pmatulis: It would be:

"This is how to access the logs ..."

docs/how-to/logs.rst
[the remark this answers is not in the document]
pmatulis: Okay, let's go with the original.

docs/how-to/logs.rst
pmatulis: We should list the other debug levels.

docs/how-to/logs.rst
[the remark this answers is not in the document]
mkalcok: Oh, of course, I didn't realize that it'll alter the sentence like this. Thanks

docs/how-to/logs.rst
mkalcok: How would you feel about linking the upstream documentation for logging @pmatulis : https://docs.openvswitch.org/en/latest/ref/ovs-appctl.8/#logging-commands

I think we shouldn't try to re-document the upstream project

pmatulis (approved):

simondeziel (approved):

mkalcok (approved):

PR #96, opened 2023-11-05
tests/upgrade: Allow DBs to settle before post-upgrade test.

fnordahl:
The upgrade test currently performs an upgrade of all nodes in quick succession, and then immediately runs tests.  This leads to spurious errors because the database clusters may or may not have settled on a new leader after all nodes being restarted.

Use `ovsdb-client` to check that each member of database cluster is connected prior to continuning with the test.


mkalcok (approved):
LGTM

PR #97, opened 2023-11-05
[branch-22.03] handle switch and chassis services on stop/refresh

fnordahl:
The upgrade test currently performs an upgrade of all nodes in quick succession, and then immediately runs tests.  This leads to spurious errors because the database clusters may or may not have settled on a new leader after all nodes being restarted.

Use `ovsdb-client` to check that each member of database cluster is connected prior to continuning with the test.

(cherry picked from commit c6eb8d494d54bb7f92c370307f6d39274967bef6)


mkalcok (approved):
LGTM

PR #99, opened 2023-11-23
Doc: Update TLS upgrade steps to restat northd

mkalcok:
In previous MicroOVN versions, restarting `microovn.daemon` would trigger restart of `northd` service as well. This is no longer the case, so the step must be performed manually.

### DIFF
--- a/docs/how-to/tls.rst
+++ b/docs/how-to/tls.rst
@@ -139,7 +139,10 @@ plaintext to TLS:
 
 2. run ``microovn certificates regenerate-ca`` on one of the cluster members
 
-3. run ``sudo snap restart microovn.daemon`` on **all** cluster members
+3. run ``sudo snap restart microovn.daemon`` on **all** cluster members. Allow
+   commands to complete before proceeding to the next step.
+
+4. run ``sudo snap restart microovn.ovn-northd`` on **all** cluster members
 
 Once this is done, OVN API services throughout the cluster will start listening
 on TLS-secured ports. However, the process is not complete yet because OVN


docs/how-to/tls.rst
```diff
@@ -141,6 +141,13 @@ plaintext to TLS:
 
 3. run ``sudo snap restart microovn.daemon`` on **all** cluster members
 
+4. run ``sudo snap restart microovn.ovn-northd`` on **all** cluster members
+
+
+.. important::
```
pmatulis, line 147: I suggest adding the admonishment before step 4. Something like:

Do not proceed to the next step until this current step is complete.

*How can the user ensure that step 3 is complete?*

pmatulis (changes requested):

docs/how-to/tls.rst
mkalcok: > How can the user ensure that step 3 is complete?

It should be fine as long as the services are not restarted at the same time. Honestly, even if they are restarted, I think the `northd` would self-recover, but as a precaution I think it's better if they are executed in sequence

mkalcok: Maybe we could even drop the warning? Maybe I was being too dramatic :D

pmatulis: Maybe just (in step 3):

"Allow the command to complete before proceeding to the next step."

pmatulis (approved):
Thanks! :grapes:

fnordahl (approved):
LGMT, thanks!

PR #100, opened 2023-11-24
snap: Fix doc build.

fnordahl:
On non-x86/arm64 architectures the doc build currently fails due to prebuilt python wheels not being available.

Add the missing build dependencies.

Use the `python3-venv` metapackage instead of specifying the specific version.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -288,7 +288,11 @@ parts:
     plugin: nil
     source: docs/
     build-packages:
-      - python3.11-venv
+      - libpython3-dev
+      - libxml2-dev
+      - libxslt1-dev
+      - make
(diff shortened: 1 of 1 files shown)


mkalcok (approved):
Makes sense to me.

Side note: Maybe we should try running build test on all supported topologies. Although I have no idea if we have such runners available

fnordahl:
> Makes sense to me.
> 
> Side note: Maybe we should try running build test on all supported topologies. Although I have no idea if we have such runners available

Indeed we should find some way to test on all architectures.

Even though we are using self-hosted runners I believe we are currently blocked by the lack of architecture support in the GitHub Actions runner itself ref actions/runner#2263 and actions/runner#2365.

PR #105, opened 2023-11-29
Update microcluster and LXD dependencies

masnax:
Updates the microcluster and LXD packages used in MicroOVN.

Namely on the microcluster side, this fixes a couple bugs:
 * A bug where the local trust store of nodes would become corrupted on sudden power loss or shutdown.
 * A bug where the `OnNewMember` hook (or `Refresh` in MicroOVN) would panic because the database hadn't opened yet.
 
Additionally, adds arguments to some of the action hooks so that CLI specified config can be propagated up to the daemons when they trigger the hooks.

On the LXD side the only real change here is that the CLI package now needs a struct to be instantiated.

I'm not sure how MicroOVN does releases, but it would be great if this could make it to the default/stable branches relatively soon, is that a possibility?


fnordahl:
/canonical/self-hosted-runners/run-workflows 2f9bbf4a657f43f4ac2d550ffa94dfa3daeb59f

fnordahl (changes requested):
Thank you for your great work on microcluster and for proposing required changes to microovn!

This looks good to me, but the CI fails due to lack of cryptographic commit signing [0], would you be able to sign and re-push the commits [1]?

0: https://github.com/canonical/microovn/actions/runs/7026800394/job/19125208812
1: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

> I'm not sure how MicroOVN does releases, but it would be great if this could make it to the default/stable branches relatively soon, is that a possibility?

This does indeed look eligible for backport. The way to approach this is to get this PR merged on the main branch, and then propose a separate PR targeting the most recent stable branch (at the time of this writing, `branch-22.03`) with cherry-picks of the relevant commits.

Note that for the backport we would like to have a separate gomod update proposed that is as small as possible, presumably it would be sufficient to only update the `microcluster` and `lxd` versions in the backport?

(The general update to gomod is fine for the main branch so let's keep that as-is.)

fnordahl (approved):
> Sounds good. The go mod changes included here are just from
>
>     go get github.com/canonical/microcluster
>     go mod tidy
>
> So they should already be the minimal set I believe.

It does appear to update to a newer point release of a lot of other deps too which is fine for main, let's review those with some more scrutiny for the backport.

LGTM, thanks!

PR #107, opened 2023-11-29
snap: Add ovn-api & ovn-chassis content interface

masnax:
Adds a `content` type snap interface so projects like LXD and MicroCloud can hook into MicroOVN and access the underlying OVN.

This will allow snapd to automatically share `/var/snap/microovn/common/data` and `/var/snap/microovn/common/run` with LXD, or any other snap that expects the `ovn-conf` interface from the same publisher.

https://snapcraft.io/docs/content-interface

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -13,6 +13,21 @@ description: |-
 
 confinement: strict
 
+slots:
+  ovn-api:
+    interface: content
+    source:
+      read:
+        - "$SNAP_COMMON/data/pki/client-cert.pem"
+        - "$SNAP_COMMON/data/pki/client-privkey.pem"
+        - "$SNAP_COMMON/data/pki/cacert.pem"
+        - "$SNAP_COMMON/data/ovn.env"
+  ovn-chassis:
+    interface: content
+    source:
+      write:
+        - "$SNAP_COMMON/run/switch"
+
 hooks:
   install:
     plugs:


mkalcok (commented):
Thanks for this contribution Max. This has been long time needed. I'm not that familiar with best practices for content interface but I'm wondering whether we are not casting a too wide net here.
This would share database files directly, as well as local control sockets. The things I can  think of that could be of interest for sharing are:
- `$SNAP_COMMON/data/pki/`
- `$SNAP_COMMON/data/ovn.env`

I might be overthinking this.

masnax:
> Thanks for this contribution Max. This has been long time needed. I'm not that familiar with best practices for content interface but I'm wondering whether we are not casting a too wide net here. This would share database files directly, as well as local control sockets. The things I can think of that could be of interest for sharing are:
> 
>     * `$SNAP_COMMON/data/pki/`
> 
>     * `$SNAP_COMMON/data/ovn.env`
> 
> 
> I might be overthinking this.

Limiting to a smaller subset is fine by me. One thing is that LXD does also grab `/run/switch` so would it be possible to include that as well?

mkalcok:
> One thing is that LXD does also grab /run/switch

Oh, I didn't know that LXD uses stuff from "runtime" directories directly.
Whatever is needed for the integration is fine by me. My only point was to not unnecessarily expose stuff that is not used.

mkalcok:
Right, I didn't realize that LXD talks directly to the controller to set external IDs and such. Perhaps I  was overthinking this after all.

mkalcok (dismissed):

fnordahl (commented):
The content interface is like an API, whatever we expose in it, needs to be supported forever. So I agree with @mkalcok that we ought to limit the exposure to the specific bits needed.

We also need to take into account other potential consumers, such as Sunbeam, k8s and others.

For local unix socket connections direct access to `run` is indeed needed.

For anything connecting to the OVN databases we would also need to expose endpoint addresses and certificate information in some way.

I'm also a bit curious about the use of `read` permission, would that really allow bi-directional communication over a unix socket for a binary in consuming snap without it also having explicit permission through other snap interfaces?

To be pragmatic though, I understand that MicroCloud has some level of urgency for this to resolve intermittent install ordering issues, so perhaps we could introduce a temporary interface to resolve that (with a name discouraging other use)?

masnax:
@mkalcok @fnordahl Sorry about the delay, this should be up to date with your suggestions now:
```
slots:
  ovn-api:
    interface: content
    source:
      read:
        - "$SNAP_COMMON/data/pki/client-cert.pem"
        - "$SNAP_COMMON/data/pki/client-privkey.pem"
        - "$SNAP_COMMON/data/pki/cacert.pem"
        - "$SNAP_COMMON/data/ovn.env"
  ovn-chassis:
    interface: content
    source:
      write:
        - "$SNAP_COMMON/run/switch"
```

mkalcok:
/canonical/self-hosted-runners/run-workflows 7d0fee101482a3de5da96be83d6d1cb851e6553c

mkalcok (approved):
Thank you @masnax This looks good to me and if it satisfies the LXD requirements for content interface I think we can merge it.

PR #110, opened 2023-12-07
wip

fnordahl:


mkalcok (commented):
Is this something that will be explored further, or can we close this PR @fnordahl ?

fnordahl:
Thanks for revisiting it, this was an experiment, and the immediate test failures serves as evidence that this won't work (not to mention other unresolved issues).

PR #113, opened 2023-12-07
microovn/ovn: support OVS `ovn-encap-ip` configuration from external initConfig

gabrielmougard:
This is needed by: https://github.com/canonical/microcloud/pull/245

### DIFF
--- a/docs/how-to/ovn-underlay.rst
+++ b/docs/how-to/ovn-underlay.rst
@@ -1,0 +1,65 @@
+===================================
+Create  custom OVN underlay network
+===================================
+
+The underlay network is the physical network infrastructure that provides connectivity between the nodes in an OVN deployment and is responsible for carrying encapsulated traffic between OVN components through Geneve (`Generic Network Virtualization Encapsulation`) tunnels.
+This allows the virtual network traffic to be transported over the physical underlay network.
+Now, by default, MicroOVN uses the hostname of a cluster member as a Geneve endpoint to set up the underlay network, but it is also possible to use custom Geneve endpoints for the cluster members.
+
+
+Set up the underlay network
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To tell MicroOVN to use the underlay network, you need to provide the IP address of the underlay network interface on each node.
+Let us assume that we want to create a three-node OVN cluster and that each node has a dedicated interface ``eth1`` with an IP address. Let says that ``10.0.1.{2,3,4}`` are the respective addresses on the ``eth1`` interface on each node.
+You can set the underlay network IP address in the :command:`init` :
+
+.. code-block:: none
+
+   microovn init
+
+Example of the interaction:
+
+.. code-block:: none
+
+   root@micro1:~# microovn init
+   Please choose the address MicroOVN will be listening on [default=10.242.68.93]:
+   Would you like to create a new MicroOVN cluster? (yes/no) [default=no]: yes
+   Please choose a name for this system [default=micro1]:
+   Would you like to define a custom encapsulation IP address for this member? (yes/no) [default=no]: yes
+   Please enter the custom encapsulation IP address for this member: 10.0.1.2
+   Would you like to add additional servers to the cluster? (yes/no) [default=no]: yes
+   What's the name of the new MicroOVN server? (empty to exit): micro2
+   eyJzZWNyZXQiOiJmOWU1OWU0N2Q1M2E0ZjJlYTYzNWYwMzIzYTE5ZTgyMjEyMzA3ZmJmY2U5OTRiOTk3NzQ4ZTAyM2VmOGEyN2MyIiwiZmluZ2VycHJpbnQiOiJlZGY0MzEzY2ZkOWFiMTdmYWIwZTZkMmE3MWZiNGZlM2U5M2RjZTBjNzNhYTQ4NWI3ZTk2Zjk2YzBhZmZlOWU2Iiwiam9pbl9hZGRyZXNzZXMiOlsiMTAuMjQyLjY4LjkzOjY0NDMiXX0=
+   What's the name of the new MicroOVN server? (empty to exit): micro3
+   eyJzZWNyZXQiOiI5MWYzODUyZTA4ZjQyOWQxNGE2Y2JiZWI0NGNmODkyMjRjNzUzZjU1NjYzYTY3MjE5ZjZkMmVhOGM0MTdhM2YxIiwiZmluZ2VycHJpbnQiOiJlZGY0MzEzY2ZkOWFiMTdmYWIwZTZkMmE3MWZiNGZlM2U5M2RjZTBjNzNhYTQ4NWI3ZTk2Zjk2YzBhZmZlOWU2Iiwiam9pbl9hZGRyZXNzZXMiOlsiMTAuMjQyLjY4LjkzOjY0NDMiXX0=
+   What's the name of the new MicroOVN server? (empty to exit):
+
+   root@micro2:~# microovn init
+   Please choose the address MicroOVN will be listening on [default=10.242.68.13]:
+   Would you like to create a new MicroOVN cluster? (yes/no) [default=no]: no
+   Please enter your join token: eyJzZWNyZXQiOiJmOWU1OWU0N2Q1M2E0ZjJlYTYzNWYwMzIzYTE5ZTgyMjEyMzA3ZmJmY2U5OTRiOTk3NzQ4ZTAyM2VmOGEyN2MyIiwiZmluZ2VycHJpbnQiOiJlZGY0MzEzY2ZkOWFiMTdmYWIwZTZkMmE3MWZiNGZlM2U5M2RjZTBjNzNhYTQ4NWI3ZTk2Zjk2YzBhZmZlOWU2Iiwiam9pbl9hZGRyZXNzZXMiOlsiMTAuMjQyLjY4LjkzOjY0NDMiXX0=
+   Would you like to define a custom encapsulation IP address for this member? (yes/no) [default=no]: yes
+   Please enter the custom encapsulation IP address for this member: 10.0.1.3
+
+   root@micro3:~# microovn init
+   Please choose the address MicroOVN will be listening on [default=10.242.68.170]:
+   Would you like to create a new MicroOVN cluster? (yes/no) [default=no]:
+   Please enter your join token: eyJzZWNyZXQiOiI5MWYzODUyZTA4ZjQyOWQxNGE2Y2JiZWI0NGNmODkyMjRjNzUzZjU1NjYzYTY3MjE5ZjZkMmVhOGM0MTdhM2YxIiwiZmluZ2VycHJpbnQiOiJlZGY0MzEzY2ZkOWFiMTdmYWIwZTZkMmE3MWZiNGZlM2U5M2RjZTBjNzNhYTQ4NWI3ZTk2Zjk2YzBhZmZlOWU2Iiwiam9pbl9hZGRyZXNzZXMiOlsiMTAuMjQyLjY4LjkzOjY0NDMiXX0=
+   Would you like to define a custom encapsulation IP address for this member? (yes/no) [default=no]: yes
+   Please enter the custom encapsulation IP address for this member: 10.0.1.4
+
+Now, the MicroOVN cluster is configured to use the underlay network with the IP addresses ``10.0.1.{2,3,4}`` on each node as tunnel endpoint for the encapsulated traffic.
+To verify that the underlay network is correctly configured, you can check the IP of OVN Geneve tunnel endpoint on each node:
+
+.. code-block:: none
+
+   root@micro1:~# microovn.ovs-vsctl get Open_vSwitch . external_ids:ovn-encap-ip
+   "10.0.1.2"
+
+   root@micro2:~# microovn.ovs-vsctl get Open_vSwitch . external_ids:ovn-encap-ip
+   "10.0.1.3"
+
+   root@micro3:~# microovn.ovs-vsctl get Open_vSwitch . external_ids:ovn-encap-ip
+   "10.0.1.4"
+
--- a/microovn/cmd/microovn/init.go
+++ b/microovn/cmd/microovn/init.go
@@ -7,8 +7,11 @@ import (
 	"os"
 
 	"github.com/canonical/lxd/lxd/util"
+	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/lxd/shared/validate"
 	"github.com/canonical/microcluster/microcluster"
+	microovnAPI "github.com/canonical/microovn/microovn/api"
 	"github.com/spf13/cobra"
 )
 
@@ -28,6 +31,24 @@ func (c *cmdInit) Command() *cobra.Command {
 	}
 
 	return cmd
+}
+
+func (c *cmdInit) wantsCustomEncapsulationIP() (string, string, error) {
+	wantsCustomEncapsulationIP, err := c.common.asker.AskBool("Would you like to define a custom encapsulation IP address for this member? (yes/no) [default=no]: ", "no")
+	if err != nil {
+		return "", "", err
+	}
+
+	if wantsCustomEncapsulationIP {
+		encapIP, err := c.common.asker.AskString("Please enter the custom encapsulation IP address for this member: ", "", validate.Required(validate.IsNetworkAddress))
+		if err != nil {
+			return "", "", err
+		}
+
+		return "ovn-encap-ip", encapIP, nil
+	}
+
+	return "", "", nil
 }
 
 func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
@@ -51,6 +72,7 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 
 	// User interaction.
 	mode := "existing"
+	customEncapsulationIpSupported := shared.ValueInSlice("custom_encapsulation_ip", microovnAPI.Extensions())
 
 	if isUninitialized {
 		// Get system name.
@@ -72,6 +94,7 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
+		var optionalConfig map[string]string
 		if wantsBootstrap {
 			mode = "bootstrap"
 
@@ -81,8 +104,19 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
+			if customEncapsulationIpSupported {
+				key, encapIP, err := c.wantsCustomEncapsulationIP()
+				if err != nil {
+					return err
+				}
+
+				if key != "" && encapIP != "" {
+					optionalConfig = map[string]string{key: encapIP}
+				}
+			}
+
 			// Bootstrap the cluster.
-			err = m.NewCluster(context.Background(), hostName, address, nil)
+			err = m.NewCluster(context.Background(), hostName, address, optionalConfig)
 			if err != nil {
 				return err
 			}
@@ -94,7 +128,21 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			err = m.JoinCluster(context.Background(), hostName, address, token, nil)
+			if customEncapsulationIpSupported {
+				// Register a potential custom encapsulation IP for other systems,
+				// so that when they will join the cluster, their encapsulation IP
+				// for the Geneve tunnel will be automatically configured.
+				key, encapIP, err := c.wantsCustomEncapsulationIP()
+				if err != nil {
+					return err
+				}
+
+				if key != "" && encapIP != "" {
+					optionalConfig = map[string]string{key: encapIP}
+				}
+			}
+
+			err = m.JoinCluster(context.Background(), hostName, address, token, optionalConfig)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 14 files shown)


gabrielmougard:
@fnordahl can you give a feedback on this when you have time please? Thanks.

mkalcok:
Hi @gabrielmougard. I left some comments at [your specification proposal](https://discourse.ubuntu.com/t/allowing-a-microcloud-user-to-specify-an-underlay-network-for-ovn/40939/2) which I think should be incorporated.
Point 1 and 2 should be, I think, easily incorporated.
Point 3, tests, could get tricky. Feel free to reach out if you have some questions.

gabrielmougard:
@mkalcok thanks a lot for the feedback! Working on it.

gabrielmougard:
@mkalcok I've been thinking about a potential issue: is there some kind of API extension system within MicroOVN that can be queryable by MicroCloud ? I'm just asking this because I wouldn't like to send a key-value config to a version of a deployed MicroOVN that does not support custom encapsualtion IP.. I'm obviously new to the MicroOVN codebase so it is maybe a dumb question. Sorry for that in advance.

tomponline:
@masnax maybe be able to offer some advice here too.

mkalcok:
Thanks for looping in Max @tomponline . I was just about to say that MicroOVN [does have an API](https://github.com/canonical/microovn/tree/main/microovn/api) that can be extended, but this "feature/version checking" should be probably handled uniformly across all microcloud components.

gabrielmougard:
@mkalcok Microcloud requires LXD, MicroCeph and MicroOVN. LXD has its own API to get API extensions (`GET /1.0` returns the list of extensions among many other metadata) and MicroCeph has https://github.com/canonical/microceph/blob/main/microceph/api/types/configs.go (`GET /1.0/configs` that returns the value of a specified key). Could we then have something similar to this KV MicroCeph API here ? Plus, you already have a compatible DB schema for supporing this:
```golang
package database

type ConfigItem struct {
	ID    int
	Key   string `db:"primary=yes"`
	Value string
}
```

It is said to be used for tracking OVN configuration but could it be also used to keep track of currently used MicroOVN features ?

gabrielmougard:
@mkalcok I don't understand why this gh runner fails for me.. Is there something I need to write in my commits for them to be accepted?

mkalcok:
It's not your fault @gabrielmougard , PRs coming users that don't have write permissions on the repository need to be manually authorised.

/canonical/self-hosted-runners/run-workflows 4a0e6f12e307e08dd334ee0499aec62cf2775b35

mkalcok:
@gabrielmougard regarding errors in the CI, I see similar issues popping up intermittently in my local environment as well . It does not look like a code issue to me, I'll look into it

mkalcok:
/canonical/self-hosted-runners/run-workflows d87cc347a1123b8a2beb5156d95a8b30aa4f9837

microovn/ovn/join.go
```diff
@@ -140,13 +142,60 @@ func Join(s *state.State, initConfig map[string]string) error {
...
+
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("Failed to get custom encapsulation IP address from database: %w", err)
+	}
+
+	// A custom encapsulation IP address can also be directly passed as an initConfig parameter.
+	// This is typically the case when an other component relying on MicroOVN (e.g, MicroCloud)
```
mkalcok, line 177: nit: These comments are very helpfull but tather than mentioning downstream projects, I think it would be better to make it clearer that the block above (pulling the encapsulation IP from config DB) is taken when user does plain `microovn cluster join` and below is the more configurable way when user uses `microovn cluster init`

mkalcok (changes requested):
Overall this looks fine,thank you for this feature. I have two small nit picks that I mentioned in the comments.
Other than that, could you please add a separate test suite that would initialize the cluster with custom IP encapsulation and then test that it was properly set?

microovn/cmd/microovn/init.go
```diff
@@ -80,8 +100,18 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
...
+			// Handle custom encapsulation IP for this system at bootstrap time.
+			key, encapIP, err := c.wantsCustomEncapsulationIP(hostName)
+			if err != nil {
+				return err
+			}
+
+			if key != "" && encapIP != "" {
+				optionalConfig = map[string]string{key: encapIP}
+			}
+
```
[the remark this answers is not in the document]
gabrielmougard: We need the `hostName` variable that might be changed in the `if` statement to be passed to `c.wantsCustomEncapsulationIP(hostName)` to further check for `key` / `encapIP`. Putting it before could lead to inconsistencies if the user decides to choose a non-default hostname.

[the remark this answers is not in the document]
mkalcok: oh, my bad. I didn't notice the change to the `hostname` arg value.

gabrielmougard:
@mkalcok is it possible to re-run the CI tests?

mkalcok:
/canonical/self-hosted-runners/run-workflows fbdd661ee3ee1331c7d3dfe8fb01c3d42da4a001

mkalcok (changes requested):
Thanks for the update @gabrielmougard. I think we need just few more touch-ups and we are done with this one. Aside from linting errors in tests, there's a way to cut down on repetition in tests themselves.

BATS doesn't really have a "native" way to load/import test suits, but we did come up with a way to do that in order to not copy/paste tests between different test suites. Example of how to do this would be [`scaleup_cluster.bats` tests](https://github.com/canonical/microovn/blob/71e939c04f9f8f9726758389efd29453873ec694/tests/scaleup_cluster.bats#L67-L73). There we execute `bats` from within the tests to "include" tests from other test suits.
What you'll need:
* custom setup/teardown (which you already have)
* way to "load" `cluster.bats` test suite as ipv4 or ipv6 variant
  * we don't have this yet but since the `cluster.bats` is already written in a way that handles ipv6, you can just create link in the `test_helper/bats/cluster_ipv6.bats -> test_helper/bats/cluster.bats`
* your test implementation (`tests/init_cluster_custom_encap.bats`) that will:
  * use your custom setup/teardown
  * tests that encapsulation is properly set in the OVN
  * loads `test_helper/bats/cluster.bats` (or `cluster_ipv6.bats`) as seen in the example above.
  * link `tests/init_cluster_custom_encap_ipv6.bats -> test/init_cluster_custom_encap.bats` to get ipv6 variant of the test suite.

masnax:
@gabrielmougard I haven't been keeping up with this too much, but I don't think the `features` endpoint should be in MicroOVN directly. Is there anything specific about this endpoint that wouldn't apply to all microcluster projects? 

Unless I'm missing something, I think all microcluster-backend projects will have API extensions and that grow over time and will need to be tracked like this, so I don't think it makes sense to re-implement this per project. In fact microcluster itself might eventually need to keep track of feature compatibility, and projects like MicroOVN would then need to manage two different implementations of this.

I think this should be an endpoint internal to microcluster, and similar to how MicroOVN passes schema extensions to microcluster, it should pass API extensions as well, which are checked when joining a cluster or upgrading an existing one. I think this way, it keeps MicroOVN focused on OVN and microcluster focused on management of the REST API and database.

gabrielmougard:
@mkalcok is it possible to re-run the CI pipeline ?

gabrielmougard:
> @gabrielmougard I haven't been keeping up with this too much, but I don't think the `features` endpoint should be in MicroOVN directly. Is there anything specific about this endpoint that wouldn't apply to all microcluster projects?
> 
> Unless I'm missing something, I think all microcluster-backend projects will have API extensions and that grow over time and will need to be tracked like this, so I don't think it makes sense to re-implement this per project. In fact microcluster itself might eventually need to keep track of feature compatibility, and projects like MicroOVN would then need to manage two different implementations of this.
> 
> I think this should be an endpoint internal to microcluster, and similar to how MicroOVN passes schema extensions to microcluster, it should pass API extensions as well, which are checked when joining a cluster or upgrading an existing one. I think this way, it keeps MicroOVN focused on OVN and microcluster focused on management of the REST API and database.

ok, I can try to implement that within microcluster.

mkalcok:
> @mkalcok is it possible to re-run the CI pipeline ?

@gabrielmougard Sorry, CI is sstill affected by [the snapcraft issue](https://bugs.launchpad.net/snapcraft/+bug/2051567), but we are looking into switching to `latest/edge` for the main branch CI, that should get this sorted out and the CI pipline gets unblocked.

mkalcok:
/canonical/self-hosted-runners/run-workflows 16090be374c9960d7d38b716d38f9a29e4c570bf

mkalcok:
@gabrielmougard CI is once again working after #119. It seems like there are still some linting issues in your PR.

If I understand correctly, you will be moving the `features` api to the `microcluster` library, correct? (I do like that approach, to have it unified across all things that use `microcluster`)

gabrielmougard:
@mkalcok can you run the CI please ?

mkalcok:
/canonical/self-hosted-runners/run-workflows 6736db623d39ae8b53d5652736d2f979763e4173

mkalcok:
Hi Gabriel, sorry about the continued CI issues. We seem to be affected by another [snapcraft issue](https://bugs.launchpad.net/snapcraft/+bug/2054099). I'll re-trigger your tests once it's solved.

In the meantime, I looked over your updated tests and they look good 👍 thanks for those.

gabrielmougard:
@mkalcok I'm removing the `1.0/features` commit of this PR. I'll open a MicroCluster PR for it and will need to open an other PR in MicroOVN to populate the the OVN related extensions at runtime.

gabrielmougard:
@mkalcok I never saw this error before:

```
 Please choose a name for this system [default=microovn-scaleup-cluster-ipv6-1]: cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# cat: /var/snap/microovn/common/run/ovn/ovn-controller.pid: No such file or directory
# stat: cannot statx '/proc/microovn_get_service_pid:': No such file or directory
```
Shall we re-run the CI or is it related to the changes in this PR?

mkalcok:
It seems to me that your `expect`ations in `microovn_init_create_cluster` do not match the order in which user prompts are displayed.

Expected:
```
Please choose the address MicroOVN will be listening on
Would you like to create a new MicroOVN cluster?
Please choose a name for this system
Would you like to define a custom encapsulation IP address for $container?
```

Actual:
```
Please choose the address MicroOVN will be listening on
Would you like to create a new MicroOVN cluster?
Would you like to define a custom encapsulation IP address for $container?
Please enter the custom encapsulation IP address for $container
Please choose a name for this system
```

gabrielmougard:
Ah! My bad. Thanks Martin!

gabrielmougard:
@mkalcok updated. But now I see:
```
#  from function `cluster_test_db_connection_string' in test file tests/init_cluster.bats, line 206)
#   `assert_output -p "ssl:${expected_addr}:${expected_port}"' failed
# Error: Failed listing services: Database is not yet initialized
# Error: Failed listing services: Database is not yet initialized
```

Is it something related to the upgrade of MicroCluster?

mkalcok:
I don't think so. Briefly looking at the CI results, it doesn't look like the cluster was properly formed.

gabrielmougard:
@mkalcok it looks better now :) There is still the `ovsdb_schema_upgrade.bats` that is failing but it does so regarless of this PR (tried the CI test on the main branch and saw that happening anyway)

gabrielmougard:
@mkalcok I also opened  this: https://github.com/canonical/microovn/pull/147 that will be useful for the OVN underlay integration in MicroCloud

microovn/ovn/join.go
```diff
@@ -142,13 +142,58 @@ func Join(s *state.State, initConfig map[string]string) error {
...
+			if err.Error() == "ConfigItem not found" {
+				return nil
+			}
+
+			return err
+		}
+
+		ovnEncapIp = item.Value
+
+		// Remove the config item as we only need it one time for the OVS configuration.
```
mkalcok, line 163: I'm bit wary of removing the key from the database right away. Are there any benefits in removing it? If there are, I'd suggest removing it as a last step in the `Join` function, otherwise if the `Join` fails after this transaction block, it can not be retried.

mkalcok: Do I understand correctly that you will remove the `DeleteConfigItem` then? 'cause I got re-request for review, but I don't see any changes :D

gabrielmougard: Ah no this was just a follow up statement.. sorry for that. But yes, I'll remove the delete operation

gabrielmougard: fixed

gabrielmougard:
@mkalcok updated

microovn/cmd/microovn/init.go
```diff
@@ -127,6 +162,23 @@ func (c *cmdInit) Run(cmd *cobra.Command, args []string) error {
...
+					// Register a potential custom encapsulation IP for other systems,
+					// so that when they will join the cluster, their encapsulation IP
+					// for the Geneve tunnel will be automatically configured.
+					key, encapIP, err := c.wantsCustomEncapsulationIP(tokenName)
+					if err != nil {
+						return err
+					}
+
+					if key != "" && encapIP != "" {
+						_, _, err = m.SQL(context.Background(), fmt.Sprintf("INSERT INTO config (key, value) VALUES ('%s', '%s')", key, encapIP))
```
mkalcok, line 175: Sorry I missed this yesterday, but wouldn't it be cleaner to use one of the generated DB mapping functions (either `CreateConfigItem` or `UpdateConfigItem`) instead of raw SQL statememnt?

mkalcok (changes requested):
Overall LGTM, I have just one more suggestion that I overlooked yesterday.

microovn/cmd/microovn/init.go
gabrielmougard: We don't have access to the MicroCluster daemon's state here, we only have access to a client (hence the use of the `m.SQL` function that is forwarding the SQL statement to the daemon) and `CreateConfigItem` requireyou to have access to the instance of the daemon State in order to create the transaction with `CreateConfigItem` inside.

mseralessandri: Can this PR https://github.com/canonical/microcluster/pull/144 be useful?

gabrielmougard: @mseralessandri I don't think this will help. Even with this PR, the state will remain internal to MicroCluster. To interact with the MicroCluster daemon holding the state, we can either:

* Define hooks that are passed to the MicroCluster app when its started.
* Can the MicroCluster API using a local client (what we do here)

As we need to have a hold on the state to create an SQL transaction in order to call `CreateConfigItem`, this is not doable (as of the current implementation of MicroCluster)

masnax: I strongly disagree with using the `SQL` command like this, that should really only be used directly from the command line, it shouldn't be incorporated into other commands.

I think what you might need if you want to do this on the init node is an API endpoint to reach out to over the unix socket that takes a payload of config and applies it on the daemon using a transaction with the pre-defined SQL mappings.

But mainly, the thing that strikes me as most odd here is that you are making a change to the database with the "expectation" that those tokens will be used, but what happens if they are not? Will the data remain in the database? If that's the case, then the next time this same node tries to join, this block would fail because the key exists and we try to create it again, no?

I feel like this whole interactive step might be better to be supplied on the joiner side and the answers should be incorporated into map that is supplied to the `PreJoin` or `PostJoin` hooks.

tomponline: Also, does this introduce the risk of SQL injection? I see that query bindings are not being used.

mkalcok:
@gabrielmougard Yes, we seems to be hitting some issue when building snap with `build-base: devel` in `snapcraft.yaml`. Without that line the build works fine. If we won't find a quick solution to this we'll temporarily remove it from the `snapcraft.yaml` in the `main` branch to unblock the CI.

gabrielmougard:
@mkalcok rebased

mkalcok:
Implementation LGTM, thank you @gabrielmougard. One last think though (and sorry for not mentioning this earlier), user facing feature like this requires also a documentation, probably a `how-to` page.

gabrielmougard:
@mkalcok I added the howto's . Let me know if this is correct ;)

docs/how-to/ovn-underlay.rst
```diff
@@ -0,0 +1,64 @@
+================================
+Creating an OVN underlay network
```
mkalcok, line 2: I realize that we don't have documented anywhere, however as per https://github.com/canonical/microovn/pull/126#issuecomment-1984172113, we are avoiding gerunds in page titles and section headers. (`Creating an OVN underlay network` -> `Create an OVN underlay network`)

docs/how-to/ovn-underlay.rst
```diff
@@ -0,0 +1,64 @@
...
+================================
+
+It is possible to create an OVN underlay network using MicroOVN. The underlay network is the physical network infrastructure that provides connectivity between the nodes in an OVN deployment.
+This underlay networks relies on the OVN Geneve (`Generic Network Virtualization Encapsulation`) tunnel, and is responsible for carrying encapsulated traffic between OVN components.
+This allows the virtual network traffic to be transported over the physical underlay network.
+
+Setting up the underlay network
```
mkalcok, line 9: same as above, no gerunds.

docs/how-to/ovn-underlay.rst
```diff
@@ -0,0 +1,64 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To tell MicroOVN to use the underlay network, you need to provide the IP address of the underlay network interface on each node.
+Let us assume that we want to create a three-node OVN cluster and that each node has a dedicated interface ``eth1`` with an IP address.Let says that ``10.0.1.<2,3,4>`` are the respective addresses on the ``eth1`` interface on each node.
```
mkalcok, line 13: nit: this is just a personal suggestion, but I would suggest using brace expansion for enumerating the IPs, ``10.0.1.{2,3,4}``

nit2: space after comma `... address.Let ...`

mkalcok (changes requested):
Overall great job on explaining underlay networks, however I think this page is slightly misleading. Reading it at face value, I'd assume that unless I use `microovn init` and set "custom encapsulation IP address", there won't be any underlay networking, which is not the case. MicroOVN, by default uses, hostname of the machine as Geneve endpoint.

Could you please add information about the default behavior and highlight that the `custom encapsulation IP` feature of `microovn init` is useful if user wants more control over which IP/interface is used for underlay, or if hostname resolution does not work between the hosts.

docs/how-to/ovn-underlay.rst
gabrielmougard: fixed

docs/how-to/ovn-underlay.rst
gabrielmougard: fixed

microovn/cmd/microovn/init.go
gabrielmougard: fixed

mkalcok (approved):
LGTM, thank you for the update and all your hard work on this one.

PR #115, opened 2024-01-25
tests: fix missing wait in TLS test suite

mkalcok:
TLS test suite setup was missing call to `wait_containers_ready` that other test suites utilize. It sometimes caused the tests to error out with:

error: too early for operation, device not yet seeded or device model not acknowledged


mkalcok:
@fnordahl This PR was rebased on top #119 and is ready for review

fnordahl (approved):
LGTM, thanks!

PR #116, opened 2024-01-25
[branch-22.03] Backport snap content interfaces

masnax:
Backports using snap content interfaces to share ovn-specific information with other snaps (namely LXD) from #107  into branch-22.03

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -12,6 +12,21 @@ description: |-
 
 confinement: strict
 
+slots:
+  ovn-api:
+    interface: content
+    source:
+      read:
+        - "$SNAP_COMMON/data/pki/client-cert.pem"
+        - "$SNAP_COMMON/data/pki/client-privkey.pem"
+        - "$SNAP_COMMON/data/pki/cacert.pem"
+        - "$SNAP_COMMON/data/ovn.env"
+  ovn-chassis:
+    interface: content
+    source:
+      write:
+        - "$SNAP_COMMON/run/switch"
+
 hooks:
   install:
     plugs:


mkalcok:
/canonical/self-hosted-runners/run-workflows 6ae7394c4a3e95cec2f6e1464dfc51fb615d4cc4

snap/snapcraft.yaml
```diff
@@ -12,6 +12,21 @@ description: |-
...
+slots:
+  ovn-api:
+    interface: content
+    source:
+      read:
+        - "$SNAP_COMMON/data/pki/client-cert.pem"
+        - "$SNAP_COMMON/data/pki/client-privkey.pem"
+        - "$SNAP_COMMON/data/pki/cacert.pem"
+        - "$SNAP_COMMON/data/ovn.env"
+  ovn-chassis:
```
fnordahl, line 24: minor note about a slight disconnect in the name of this slot promising access to both OVS and OVN sockets, and that it in this point in time only provides the OVS sockets.

I assume this is because only the OVS sockets are needed. If more is needed later we can always extend/move things.

fnordahl (approved):
LGTM, thanks!

snap/snapcraft.yaml
masnax: If you think it would be more correct, I can rename this to `ovs-chassis` and `ovn-chassis` can be introduced later if it becomes necessary.

fnordahl: No need to change now, just wanted to make a note of it.

fnordahl: No need to change now, just wanted to make a note of it.

PR #117, opened 2024-01-26
build: set microovn's binary version at build time

mkalcok:
This fixes incorrect version (`0.1`) reported by `microovn --version`.
New behavior is to output current version of `MicroOVN` along with
`ovn` and `openvswitch` versions of packages from which it was built.

Closes LP#2049941

---

This change was inspired by [the way the MicroCeph sets its version](https://github.com/sabaini/microceph/blob/70e7b70031059aec85d24ade3a329ec79c7b1eb0/snap/snapcraft.yaml#L299-L307).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -4,7 +4,7 @@ base: core24
 build-base: devel
 assumes:
  - snapd2.59
-adopt-info: ovn
+adopt-info: microovn
 grade: devel
 source-code: https://github.com/canonical/microovn.git
 summary: Simple clustered OVN deployment
@@ -223,18 +223,6 @@ parts:
         craftctl default
 
         mkdir -p "${CRAFT_PART_INSTALL}/etc/ovn/"
-    override-stage: |
-        craftctl default
-        pkg_version=$(
-            dpkg-deb -f \
-            $CRAFT_PART_SRC/../stage_packages/ovn-host*.deb \
-            Version | sed -rne 's/([0-9.]+)[-+~].*$$/\1/p')
-        git_version=$(
-            git -C $CRAFT_PROJECT_DIR describe \
-                --always \
-                --dirty \
-                --abbrev=10)
-        craftctl set version=${pkg_version}+snap${git_version}
 
   ovs:
     plugin: nil
@@ -275,6 +263,8 @@ parts:
     source: microovn/
     after:
       - dqlite
+      - ovn
+      - ovs
     build-snaps:
       - go
     plugin: nil
@@ -292,8 +282,35 @@ parts:
       export CGO_LDFLAGS="-L${CRAFT_STAGE}/lib/ -L${CRAFT_STAGE}/usr/local/lib/"
       export CGO_LDFLAGS_ALLOW="(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"
 
+      # Determine versions of packages used to build MicroOVN
+      # Note (mkalcok): This code reaches into other snapcraft parts via hardcoded paths
+      #                 to determine version of packages from which they were built. It
+      #                 will require updates if these part names ever change.
+      ovn_pkg_version=$(
+          dpkg-deb -f \
+          $CRAFT_PART_SRC/../../ovn/stage_packages/ovn-host*.deb Version)
+      ovs_pkg_version=$(
+          dpkg-deb -f \
+          $CRAFT_PART_SRC/../../ovs/stage_packages/openvswitch-switch*.deb Version)
+      git_version=$(
+          git -C $CRAFT_PROJECT_DIR describe \
+              --always \
+              --dirty \
+              --abbrev=10)
+      
+      # Set Snap's version string
+      # Note (mkalcok): Short version string does not include potential package snapshot hash.
+      #                 This is required because Snap's version string is limited to 32 chars
+      #                 and inclusion of snapshot hash violates this boundary.
+      ovn_pkg_short_version=$(echo $ovn_pkg_version | sed -rne 's/([0-9.]+)[-+~].*$$/\1/p')
+      craftctl set version=${ovn_pkg_short_version}+snap${git_version}
+      
       # Build the binaries
-      go build -o "${CRAFT_PART_INSTALL}/bin/microovn" ./cmd/microovn
+      go build -o "${CRAFT_PART_INSTALL}/bin/microovn" \
+               -ldflags "-X 'main.MicroOvnVersion=${git_version}' \
+                         -X 'main.OvnVersion=${ovn_pkg_version}' \
+                         -X 'main.OvsVersion=${ovs_pkg_version}'" \
+               ./cmd/microovn
       go build -o "${CRAFT_PART_INSTALL}/bin/microovnd" -tags=libsqlite3 ./cmd/microovnd
     prime:
       - bin/microovn
--- a/microovn/cmd/microovn/main.go
+++ b/microovn/cmd/microovn/main.go
@@ -3,6 +3,7 @@ package main
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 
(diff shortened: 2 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -226,9 +226,10 @@ parts:
     override-stage: |
         craftctl default
         pkg_version=$(
-            dpkg-deb -f \
-            $CRAFT_PART_SRC/../stage_packages/ovn-host*.deb \
-            Version | sed -rne 's/([0-9.]+)[-+].*$$/\1/p')
+            apt-cache policy ovn-host | \
```
fnordahl, line 229: The current code is getting the version from the actual binary consumed by the build, the new version gets the version from some cache, which may or may not be in sync with the actually binary used.

Why do you think this change is necessary, can we keep the existing method?

snap/snapcraft.yaml
```diff
@@ -292,8 +293,22 @@ parts:
       export CGO_LDFLAGS="-L${CRAFT_STAGE}/lib/ -L${CRAFT_STAGE}/usr/local/lib/"
       export CGO_LDFLAGS_ALLOW="(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"
 
+      # Determine OVN package version
+      pkg_version=$(
+          apt-cache policy ovn-host | \
```
fnordahl, line 298: See comment above, I'd rather et this from the actual binary consumed by the build.

snap/snapcraft.yaml
```diff
@@ -292,8 +293,22 @@ parts:
...
+          awk '/Candidate:/{ print $2 }' | \
+          sed -rne 's/([0-9.]+)[-+].*$$/\1/p'
+      )
+      git_version=$(
```
fnordahl, line 302: Instead of doing all this twice, should we perhaps instead make the `microovn` part build after `ovn`, and point `adopt-info` to the `microovn` part?

fnordahl (changes requested):
I have some comments on the approach in-line.

I also wonder if we should take the opportunity to make a larger change to this since the embedded version string appears to not be used by anyone.

We have been discussing the need to expose versions of other components not currently recorded as part of the version string, such as the version of Open vSwitch.

I think it could be usefull if a call to `microovn --version` produced this information?

Something like:
```
$ microovn --version
microovn 1d18f95c73
ovn 22.03.3-0ubuntu0.22.04.1
openvswitch 2.17.8-0ubuntu0.22.04.1
...
```

mkalcok:
Thank you for the review @fnordahl. I'll address your inline comments below as they are all connected.

+1 on including `ovs` version in the version string as well.

Why to use `apt-cache` to determine `ovn` version in `microovn` part?
The `ovn` part was using `dpkg-deb` to get version directly form the `.deb` package because it had easy way to point to the file via `$CRAFT_PART_SRC` variable. The same is not true for `microovn` part, we'd have to hard-code path for the `.deb` file which would be more prone to breaking in the future.

Why to use `apt-cache` to determine `ovn` version in `ovn` part?
I changed to `ovn` part to be in line with `microovn` part, so that the two parts don't use different methods of fetching the `ovn` version.

Can the `apt-cache` output report inaccurate version?
I don't think so. We use `stage-packages` in `ovn` part to pull `ovn-host` package. So whether we look at the pulled `.deb` file for the version or we ask the system that provided the `.deb` file, I think the reported version should be the same.

Why not use `adopt-info`?
I wasn't aware of this snapcraft feature. I took a quick look and I'm not sure I completely understand how we can use it to pass data between parts, but I'll definitely look into this more. If there's a way to determine ovn/ovs/microovn version only once, we should definitely go for it.

fnordahl:
> Why to use `apt-cache` to determine `ovn` version in `microovn` part?
> The `ovn` part was using `dpkg-deb` to get version directly form the `.deb` package because it had easy way to point to the file via `$CRAFT_PART_SRC` variable. The same is not true for `microovn` part, we'd have to hard-code path for the `.deb` file which would be more prone to breaking in the future.

Yeah, I can see how reaching into other parts files is a bit of a boundary violation, I still like getting information from actual binary better than `apt-cache` though. I can envision many potential cases where subsequent runs on the same build environment could potentially lead to difference in state between cache and downloaded binary.

Any other way we could extract version information from the various parts for subsequent collection in the `microovn` part?

> Why not use `adopt-info`?
> I wasn't aware of this snapcraft feature. I took a quick look and I'm not sure I completely understand how we can use it to pass data between parts, but I'll definitely look into this more. If there's a way to determine ovn/ovs/microovn version only once, we should definitely go for it.

It basically tells snapcraft which part will be responsible for making the `craftctl set version` call.

mkalcok:
> Any other way we could extract version information from the various parts for subsequent collection in the microovn part?

What if  both `ovn` and `ovs` parts created a file `<part_name>.version` in `$CRAFT_PROJECT_DIR` or `$CRAFT_PRIME` and then the `microovn` part would just read this files to construct the final version string?

fnordahl:
> What if both `ovn` and `ovs` parts created a file `<part_name>.version` in `$CRAFT_PROJECT_DIR` or `$CRAFT_PRIME` and then the `microovn` part would just read this files to construct the final version string?

Creating a file sounds reasonable, using `$CRAFT_PROJECT_DIR` might work but could be susceptible to being overwritten when snapcraft resyncs the project dir from the executing host which may happen independently of the steps for the `ovn` and `ovs` parts, not sure.

I'd avoid `$CRAFT_PRIME` as we might end up getting the artifacts into the snap.

Since we will need to build the name of the file by statically referencing part name anyway, is reaching into the parts build/work directory really that bad?

mkalcok:
You are probably right, creating arbitrary files in the snap build env could cause unforeseen issues in repeated builds. Using hardcoded part names will be probably fine because any issues will manifest at build time.

snap/snapcraft.yaml
mkalcok: Oh, I totally misunderstood the meaning of this comment the first time i read it. I see what you mean, you want to move setting of the snap's version to this part instead of `ovn`.
Yeah, that make's sense. I'll mark this as WIP until I apply this change.

mkalcok: Updated. Snap's version is now set in `microovn` part.

fnordahl:
> I applied your suggestions @fnordahl and I'm sourcing the package version from `dpkg-deb` output, similar to the way `ovn` part sets the snap's version. One notable difference though is that since there are no limits on character length, I'm including the potential git hash of snapshoted package as well. The output looks like this:
> 
> ```
> # microovn --version
> microovn: e0d7df4d56-dirty
> ovn: 24.03.0~git20240205.ebb7076
> openvswitch: 3.3.0~git20240118.e802fe7
> ```

Nice, both the approach and result looks great!

I wonder however if we should include both the upstream and downstream version for ovs/ovn, i.e. retain the entire package version string. That way this becomes a sort of manifest where it would be possible to quickly identify the package patch level used in the snap.

While ancient history now, [the diff between the most recent 20.03 package and the previous](http://launchpadlibrarian.net/613956279/ovn_20.03.2-0ubuntu0.20.04.3_20.03.2-0ubuntu0.20.04.4.diff.gz) is an example of a significant downstream patch delta.

mkalcok:
Yup, it makes sense to include complete package version. I updated the proposal and currently the output looks like this:

```
# microovn --version
microovn: 131e78b880-dirty
ovn: 24.03.0~git20240205.ebb7076-1
openvswitch: 3.3.0-1
```

fnordahl (approved):
LGTM, thanks!

fnordahl:
@mkalcok I'd merge this now, but it appears a manual rebase is needed.

mkalcok:
I re-triggered tests because it seemed like an CI issue, something about getting `503` when fetching snap from snapstore.

PR #119, opened 2024-02-13
CI: Install snapcraft from latest/edge on main branch.

fnordahl:
The main branch tracks the current Ubuntu development release, this gives us early warning on coming issues for future releases which we can use both to prepare and to feed issues to projects we depend on early.

We now have a concrete example where builds are blocked and a fix is available on the snapcraft latest/edge channel.

Use the latest/edge channel for snapcraft on main branch, with fallback to the latest/stable channel on other branches.

The fallback minimizes the manual work required when branching for future stable releases.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,6 +26,15 @@ jobs:
     runs-on: [self-hosted, linux, X64, jammy, large]
     env:
       MICROOVN_SNAP: microovn.snap
+      # The `base_ref` will only be set for PR and contain the name of the
+      # target branch.  The `ref_name` will be correct for the final push
+      # check after a PR is merged.
+      #
+      # This setup may lead to failures on push to arbitrarily named branches
+      # on a fork, but that is a price worth paying.
+      #
+      # Contributors can raise a draft PR to get accurate results.
+      POSSIBLE_TARGET_BRANCH: "${{ github.base_ref || github.ref_name }}"
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


mkalcok:
The tests fail on setting "too long" snap version. Is this because the deb package, from which we parse the final version,  contains git hash?

fnordahl:
> The tests fail on setting "too long" snap version. Is this because the deb package, from which we parse the final version, contains git hash?

This is a separate issue that we also need to address, I guess it makes sense to do it as part of this PR.

We will routinely be uploading snapshots of OVS and OVN to Ubuntu at this point in the cycle using this type of package version, both to help upstream with early exposure and testing and to help weed out issues before the Ubuntu release.

So let's look at how the version string is built and adapt the procedure to cope with these types of version strings.

mkalcok:
I'd say that with addition of correct version reported by microovn itself (https://github.com/canonical/microovn/pull/117), we can stick, to brief "24.03" for the snap version string to avoid hitting length limit. Detailed version can be then reported by the output `microovn --version`

fnordahl:
> I'd say that with addition of correct version reported by microovn itself (#117), we can stick, to brief "24.03" for the snap version string to avoid hitting length limit. Detailed version can be then reported by the output `microovn --version`

That  does however have the additional cost of having to request that information for every bug request we get, whereas now it is immediately know from the snap version string whether something changed in microovn or in Ubuntu.

Since these two repositories move independent of each other and can change as a result of a rebuild without the microovn repository having changed knowing which is which is quite important.

The specific reason for the string to overrun is known, there is no need to change the versioning scheme to fix it!

mkalcok (approved):
LGTM. Thanks a lot for unblocking our CI pipeline!

PR #123, opened 2024-02-26
snap: Use only directories for ovn-api content interface

masnax:
Sadly, as it turns out, we are not able to specify individual files to a content interface. On the consuming side, each entry will be created as a directory and populated with its contents. The current implementation is resulting in LXD seeing this, where each directory is empty:

```
snap run --shell lxd -c  "aa-exec -p unconfined ls -l /var/snap/lxd/current/microovn/api"
total 16
drwxr-xr-x 2 root root 4096 Feb 26 21:54 cacert.pem
drwxr-xr-x 2 root root 4096 Feb 26 21:54 client-cert.pem
drwxr-xr-x 2 root root 4096 Feb 26 21:54 client-privkey.pem
drwxr-xr-x 2 root root 4096 Feb 26 21:54 ovn.env

```

This isn't just the case with the LXD state directory, the actual source is also overwritten in microovn:
```
ls -l /var/snap/microovn/common/data/pki/
total 12
drwxr-xr-x 2 root root 4096 Feb 26 21:54 cacert.pem
drwxr-xr-x 2 root root 4096 Feb 26 21:54 client-cert.pem
drwxr-xr-x 2 root root 4096 Feb 26 21:54 client-privkey.pem
```

So I've had to include all of `data` into the content interface. The other option here would be to restructure the `data` directory.

### DIFF
--- a/microovn/ovn/paths/paths.go
+++ b/microovn/ovn/paths/paths.go
@@ -47,7 +47,7 @@ func SwitchDataDir() string {
 
 // OvnEnvFile returns path to the file used to configure env variables for OVN commands
 func OvnEnvFile() string {
-	return filepath.Join(dataDir, "ovn.env")
+	return filepath.Join(EnvDir(), "ovn.env")
 }
 
 // OvnNBDatabaseSock returns path to the local unix socket used by Northbound OVN database
@@ -78,6 +78,11 @@ func OvsDatabaseSock() string {
 // PkiDir returns path to the directory that store OVN certificates
 func PkiDir() string {
 	return filepath.Join(dataDir, "pki")
+}
+
+// EnvDir returns path to the directory that store OVN environment variables.
+func EnvDir() string {
+	return filepath.Join(dataDir, "env")
 }
 
 // PkiCaCertFile returns path to CA certificate file
@@ -129,6 +134,7 @@ func RequiredDirs() []string {
 		SwitchDataDir(),
 		LogsDir(),
 		PkiDir(),
+		EnvDir(),
 	}
 }
 
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -2,3 +2,9 @@
 
 mkdir -p $SNAP_COMMON/logs
 cp $SNAP/docs/microovn/how-to/logs.txt $SNAP_COMMON/logs/README
+
+# If we upgraded from snap revision 379 (24.03.0+snap395808ff84) or earlier, ensure we move ovn.env to its own directory before running the start commands.
+if test -e "${SNAP_COMMON}/data/ovn.env" ; then
+  mkdir "${SNAP_COMMON}/data/env" || true
+  mv "${SNAP_COMMON}/data/ovn.env" "${SNAP_COMMON}/data/env/ovn.env"
+fi
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,14 +14,16 @@ description: |-
 confinement: strict
 
 slots:
-  ovn-api:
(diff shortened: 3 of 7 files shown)


masnax:
This is currently causing very odd behaviour in `22.03/edge` where `microovn cluster bootstrap` occasionally complains that the certs are actually directories, which means there is some sort of race going on between microovn populating the certs and the content interface detecting them, deleting them, and replacing them with a directory of the same name.

mkalcok:
Hi @masnax thank you for looking into this. I'm still bit concerned about exposing entire `data/`. The way I see it, if we expose it as a whole, we won't be able to change it's structure in the future as it could break anything that uses this interface and relies on its specific structure.

As an alternative, what would you think about moving `ovn.env` into its own directory (`$SNAP_COMMON/data/env/`) and instead of current
```yaml
  ovn-api:
    interface: content
    source:
      read:
        - "$SNAP_COMMON/data/pki/client-cert.pem"
        - "$SNAP_COMMON/data/pki/client-privkey.pem"
        - "$SNAP_COMMON/data/pki/cacert.pem"
        - "$SNAP_COMMON/data/ovn.env"
```
We'd have
```yaml
  ovn-certificates:
    interface: content
    source:
      read:
        - "$SNAP_COMMON/data/pki/"
  ovn-env:
    interface: content
    source:
      read:
        - "$SNAP_COMMON/data/env/"
```
This would require slight change in the microovn code itself, but there are not many places where we refer to the `ovn.env` file explicitly with full path. It's mostly in tests, [microovn's path definition](https://github.com/canonical/microovn/blob/395808ff84767c1038163e363c9cb4d25cef5f5f/microovn/ovn/paths/paths.go#L49) and [command wrappers](https://github.com/canonical/microovn/tree/main/snapcraft). (Note that some command wrappers refer to `$SNAP/ovn.env`, but that's file unrelated to this change)
(cc @fnordahl )

masnax:
@mkalcok Thanks for the guidance, I've set it up as you've described.

mkalcok:
/canonical/self-hosted-runners/run-workflows 576f41f47f4c6a238bd68d37f24b932dbe40f4db

mkalcok:
Thank you for the persistence :D It looks good to me at first glance, I'll give it more thorough look/testing tomorrow.

masnax:
@mkalcok You suspected right, indeed it was the commands in `snapcraft/commands/*.start` that were attempting to source the new `ovn.env` before the daemon creates it.

To work around this, I added a case to the `post-refresh` hook to move `ovn.env` to its new home. This should also clean up the old `ovn.env` so it won't hang around the state directory, potentially confusing people: https://github.com/canonical/microovn/pull/123/commits/90d88ad509d7888233fc86ec5193b7f798c779a7

fnordahl:
/canonical/self-hosted-runners/run-workflows 90d88ad509d7888233fc86ec5193b7f798c779a7

mkalcok (approved):
Thanks for the updates. LGTM

fnordahl:
@masnax thanks a lot for your work on this. With this merged on main, a question arises for the state of branch-22.03, as it currently holds a broken version of this feature. To inform our decision on how to proceed, would you want to see this work for the current stable snap, or would you see this as a feature of the coming 24.03 stable release?

If the former, would you want to post a backport to branch-22.03?

cc @mkalcok

tomponline:
@fnordahl I think we would like to see this backported to current stable if possible.
@masnax do you concur?

masnax:
@fnordahl @tomponline Yes absolutely, that was always the intention.

masnax:
https://github.com/canonical/microovn/pull/125

PR #124, opened 2024-03-06
Improvements to Make targets and test execution

mkalcok:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ ifndef MICROOVN_SNAP_CHANNEL
 	export MICROOVN_SNAP_CHANNEL="22.03/stable"
 endif
 
+ALL_TESTS := $(wildcard tests/*.bats)
+
 check: check-lint check-system
 
 check-tabs:
@@ -17,9 +19,11 @@ check-lint: check-tabs
 		-not -name \*.swp \
 		| xargs shellcheck --severity=warning && echo Success!
 
-check-system: $(MICROOVN_SNAP)
-	echo "Running functional tests";					\
-	$(CURDIR)/.bats/bats-core/bin/bats tests/
+$(ALL_TESTS): $(MICROOVN_SNAP)
+	echo "Running functional test $@";					\
+	$(CURDIR)/.bats/bats-core/bin/bats $@
+
+check-system: $(ALL_TESTS)
 
 $(MICROOVN_SNAP):
 	echo "Building the snap";						\
@@ -28,3 +32,5 @@ $(MICROOVN_SNAP):
 clean:
 	rm -f $(MICROOVN_SNAP_PATH);						\
 	snapcraft clean
+
+.PHONY: $(ALL_TESTS) clean check-system check-lint check-tabs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,6 +5,11 @@ ifndef MICROOVN_SNAP_CHANNEL
 	export MICROOVN_SNAP_CHANNEL="22.03/stable"
 endif
 
+ALL_TESTS := $(wildcard tests/*.bats)
+MICROOVN_SOURCES := $(shell find microovn/ -type f)
```
fnordahl, line 9: Neat trick, and I guess this is OK since we're only using it for detecting whether a (re-)build is required.  We would need to explicitly list files if we want to use it for something like `make dist` or similar in the future, but I see no need for that.

Makefile
```diff
@@ -17,14 +22,18 @@ check-lint: check-tabs
...
-$(MICROOVN_SNAP):
+$(MICROOVN_SNAP): $(MICROOVN_SOURCES) $(SNAP_SOURCES) $(COMMAND_WRAPPERS)
 	echo "Building the snap";						\
 	snapcraft pack -v -o $(MICROOVN_SNAP)
 
 clean:
 	rm -f $(MICROOVN_SNAP_PATH);						\
 	snapcraft clean
+
+.PHONY: $(ALL_TESTS) clean check-system check-lint check-tabs
```
fnordahl, line 39: Do we really want to run system tests when just running `make`? I think I'd expect a plain `make` to only build the snap.

fnordahl (commented):
I have one comment about the phony target.

Just ooi, what is the tradeoff between having make do the parallelization vs. having bats do the parallelization through its `-j`, `--no-parallelize-across-files` and `--no-parallelize-within-files` arguments?

Otherwise looks like a neat change and will surely improve developer ergonomics.

Makefile
mkalcok: I may misunderstand something about `.PHONY`, but my intention was ensure that all tests in `$(ALL_TESTS)` are always executed. Since the test targets (recipes?) are now real files, `Make` could (and I saw it happen in my testing) assume that the target is "up-to-date" and not execute it.

As for adding `clean`, `check-system` and others, I included them out of habit. I think it's better to have these "virtual" targets  explicitly in the `.PHONY` on the off chance that a file will be created with the same name sometime in the future and will cause unexpected behavior of `make`.

> Do we really want to run system tests when just running make?

This is already happening with current implementation. Since the `check-system` target does not represent any real file, it's always "out-of-date" so plain `make` will re-execute it. I'll look into how to fix this behavior (I'm happy to take any suggestions, if you have some.).

mkalcok:
> Just ooi, what is the tradeoff between having make do the parallelization vs. having bats do the parallelization through its -j, --no-parallelize-across-files and --no-parallelize-within-files arguments?

Parallelization on `make` level would be equal to `--no-prallelize-within-files`, which is what we want. Added value is that it's exposed through familiar `-j` option to `make`. The only downside that I observed is that in order to get useful output from the parallelized run, you need to use [`-O`](https://www.gnu.org/software/make/manual/html_node/Parallel-Output.html). Otherwise the output of multiple tests will interleave.

Makefile
mkalcok: > Do we really want to run system tests when just running make?

I added `.DEFAULT_GOAL` so that when user runs `make`, it only builds the snap without running tests.

fnordahl (approved):
LGTM, thanks!

PR #127, opened 2024-04-30
Avoid blocking issue in snapd 2.62

mkalcok:
snapd 2.62 fails to install microovn in containers. This temporary change installs snapd from latest/beta channel to sidestep this issue in functional tests.


fnordahl (approved):
LGTM, thanks!

PR #128, opened 2024-05-15
Support daemon mode for ovn-*bctl commands.

fnordahl:
When running the ``ovn-nbctl`` and ``ovn-sbctl`` commands with the ``--detach`` argument the binary will daemonize and listen on a unix socket.

To allow this we need to add the ``network-bind`` plug.

### DIFF
--- a/tests/test_helper/bats/cli_ovsovn.bats
+++ b/tests/test_helper/bats/cli_ovsovn.bats
@@ -22,6 +22,18 @@ teardown() {
             break
         done
     fi
+
+    if [ "$BATS_TEST_DESCRIPTION" = "ovn-nbctl daemon" ]; then
+        for container in $TEST_CONTAINERS; do
+            lxc_exec "$container" "killall ovn-nbctl"
+        done
+    fi
+
+    if [ "$BATS_TEST_DESCRIPTION" = "ovn-sbctl daemon" ]; then
+        for container in $TEST_CONTAINERS; do
+            lxc_exec "$container" "killall ovn-sbctl"
+        done
+    fi
 }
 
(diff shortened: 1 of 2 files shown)


tests/test_helper/bats/cli_ovsovn.bats
```diff
@@ -22,6 +22,18 @@ teardown() {
             break
         done
     fi
+
```
mkalcok, line 25: nit: Conditional teardown based on the test name seems bit fragile. I wonder if it wouldn't be better to always execute these steps and just ignore errors

mkalcok (approved):
One small nit for the consideration. Otherwise LGTM

tests/test_helper/bats/cli_ovsovn.bats
fnordahl: Not sure if I follow, the condition is to only run them for the tests that actually start the daemons?

In any case, if the `killall` fails, it does indeed already just continue.

mkalcok: I meant that if someone renames the test in future and they are not aware that they need to change the condition in the teardown as well, it will stop performing the teardown correctly.
My suggestion was to call these `killall` commands always, regardless of the test name.

fnordahl: An approach like that has the potential to become very noisy, and for any expensive teardown functions it could easily lead to extended time spent on test runs.

An alternative approach could be to store a stack somewhere and have the tests register their teardowns instead.

Not sure if we can access the test name from within the test definition, but it would at least put the two identifiers in close proximety which could reduce the risk of what you're describing.

PR #130, opened 2024-05-28
Ovsdb schema upgrade

mkalcok:
Please review each commit individually.

This change set lets MicroOVN to take control of the OVN OVSDB cluster schema upgrades.

It also includes few refactoring changes, that are in my opinion required.  The `microovn/ovn` package got bit bloated over time and it became hard to use its functionality from other packages. Importing `microovn/ovn` would often result in circular dependencies, so I attempted to extract, and group, functions that could stand on their own into standalone packages.

I consider the implementation good for review. Functional tests and documentation are in progress.

### DIFF
--- a/microovn/api/ovsdb/schema_version.go
+++ b/microovn/api/ovsdb/schema_version.go
@@ -1,6 +1,7 @@
 package ovsdb
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/microcluster/client"
 	"github.com/canonical/microcluster/rest"
 	"github.com/canonical/microcluster/state"
 	"github.com/canonical/microovn/microovn/api/types"
@@ -33,6 +35,12 @@ var ExpectedSchemaVersion = rest.Endpoint{
 	Get:  rest.EndpointAction{Handler: getExpectedSchemaVersion, AllowUntrusted: false, ProxyTarget: false},
 }
 
+// AllExpectedSchemaVersions defines endpoints for /1.0/ovsdb/schema/<db-name>/expected/all
+var AllExpectedSchemaVersions = rest.Endpoint{
+	Path: "ovsdb/schema/{db}/expected/all",
+	Get:  rest.EndpointAction{Handler: getAllExpectedSchemaVersions, AllowUntrusted: false, ProxyTarget: false},
+}
+
 // ActiveSchemaVersion defines endpoints for /1.0/ovsdb/schema/<db-name>/active
 var ActiveSchemaVersion = rest.Endpoint{
 	Path: "ovsdb/schema/{db}/active",
@@ -56,6 +64,54 @@ func getExpectedSchemaVersion(s *state.State, r *http.Request) response.Response
 	}
 
 	return response.SyncResponse(true, expectedVersion)
+}
+
+// getAllExpectedSchemaVersions implements GET method for /1.0/ovsdb/schema/<db-name>/expected/all.
+// It returns expected schema version for the given database from each node in the deployment.
+// The response is in the format of types.OvsdbSchemaReport
+func getAllExpectedSchemaVersions(s *state.State, r *http.Request) response.Response {
+	dbSpec, errResponse := parseDbSpec(r)
+	if errResponse != nil {
+		return errResponse
+	}
+
+	// Get local expected schema version and store it in the final result
+	localExpectedVersion, err := ovsdb.ExpectedOvsdbSchemaVersion(s, dbSpec)
+	if err != nil {
+		logger.Errorf("failed to get expected OVSDB schema version for '%s' database: %s", dbSpec.FriendlyName, err)
+		return response.ErrorResponse(500, "Internal Server Error")
+	}
+
+	responseData := types.OvsdbSchemaReport{
+		types.OvsdbSchemaVersionResult{
+			SchemaVersion: localExpectedVersion,
+			Host:          s.Address().Hostname(),
+			Error:         types.OvsdbSchemaFetchErrorNone,
+		},
+	}
+
+	// Get clients for each member in the cluster
+	clusterClient, err := s.Cluster(r)
+	if err != nil {
+		logger.Errorf("Failed to get a client for every cluster member: %s", err)
+		return response.ErrorResponse(500, "Internal Server Error")
+	}
+
+	// Fetch expected schema versions from each cluster member.
+	_ = clusterClient.Query(s.Context, true, func(ctx context.Context, c *client.Client) error {
+		clientUrl := c.URL()
+		logger.Debugf("Fetching expected OVN %s schema version from '%s'", dbSpec.FriendlyName, clientUrl.String())
+		nodeStatus := types.OvsdbSchemaVersionResult{Host: clientUrl.Hostname()}
+
+		result, responseSuccess := microovnClient.GetExpectedOvsdbSchemaVersion(ctx, c, dbSpec)
+		nodeStatus.Error = responseSuccess
+		nodeStatus.SchemaVersion = result
+
+		responseData = append(responseData, nodeStatus)
+		return nil
+	})
+
+	return response.SyncResponse(true, &responseData)
 }
 
 // getActiveSchemaVersion implements GET method for /1.0/ovsdb/schema/<db-name>/active. It returns
--- a/microovn/api/ovsdb/schema_version.go
+++ b/microovn/api/ovsdb/schema_version.go
@@ -1,0 +1,64 @@
+package ovsdb
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/state"
+	ovnCmd "github.com/canonical/microovn/microovn/ovn/cmd"
+	"github.com/gorilla/mux"
+
+	"github.com/canonical/microovn/microovn/ovn/ovsdb"
+)
+
+// supportedDBs contains all database types that support reporting of
+// a schema version.
+var supportedDBs = map[string]ovnCmd.OvsdbType{
+	"sb":     ovnCmd.OvsdbTypeSBLocal,
+	"nb":     ovnCmd.OvsdbTypeNBLocal,
+	"switch": ovnCmd.OvsdbTypeSwitchLocal,
+}
+
+// ExpectedSchemaVersion defines endpoints for /1.0/ovsdb/schema/<db-name>/expected
+var ExpectedSchemaVersion = rest.Endpoint{
+	Path: "ovsdb/schema/{db}/expected",
+	Get:  rest.EndpointAction{Handler: getExpectedSchemaVersion, AllowUntrusted: false, ProxyTarget: false},
+}
+
+// getExpectedSchemaVersion implements GET method for /1.0/ovsdb/schema/<db-name>/expected. It returns
+// expected schema version of the specified database on this MicroOVN node. Expected schema version is
+// determined by looking at the schema file that is bundled with OVN/OVS packages.
+// URL variable <db-name> is expected to be one of the keys from supportedDBs.
+func getExpectedSchemaVersion(s *state.State, r *http.Request) response.Response {
+	requestedDB, err := url.PathUnescape(mux.Vars(r)["db"])
+	if err != nil {
+		logger.Errorf("failed to parse requested DB name from url '%s'", r.URL)
+		return response.ErrorResponse(500, "Internal Server Error")
+	}
+	requestedDB = strings.ToLower(requestedDB)
+
+	dbType, ok := supportedDBs[requestedDB]
+
+	if !ok {
+		return response.BadRequest(fmt.Errorf("DB '%s' not supported", requestedDB))
+	}
+
+	dbSpec, err := ovnCmd.NewOvsdbSpec(dbType)
+	if err != nil {
+		logger.Errorf("%s", err)
+		return response.ErrorResponse(500, "Internal Server Error")
+	}
+
+	expectedVersion, err := ovsdb.ExpectedOvsdbSchemaVersion(s, dbSpec)
+	if err != nil {
+		logger.Errorf("failed to get expected OVSDB schema version for '%s' database: %s", dbSpec.FriendlyName, err)
+		return response.ErrorResponse(500, "Internal Server Error")
+	}
+
+	return response.SyncResponse(true, expectedVersion)
+}
--- a/microovn/api/ovsdb/schema_version.go
+++ b/microovn/api/ovsdb/schema_version.go
@@ -10,6 +10,9 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microcluster/rest"
 	"github.com/canonical/microcluster/state"
+	"github.com/canonical/microovn/microovn/api/types"
+	microovnClient "github.com/canonical/microovn/microovn/client"
+	"github.com/canonical/microovn/microovn/node"
 	ovnCmd "github.com/canonical/microovn/microovn/ovn/cmd"
 	"github.com/gorilla/mux"
 
@@ -30,28 +33,20 @@ var ExpectedSchemaVersion = rest.Endpoint{
 	Get:  rest.EndpointAction{Handler: getExpectedSchemaVersion, AllowUntrusted: false, ProxyTarget: false},
 }
 
+// ActiveSchemaVersion defines endpoints for /1.0/ovsdb/schema/<db-name>/active
+var ActiveSchemaVersion = rest.Endpoint{
+	Path: "ovsdb/schema/{db}/active",
+	Get:  rest.EndpointAction{Handler: getActiveSchemaVersion, AllowUntrusted: false, ProxyTarget: false},
+}
+
 // getExpectedSchemaVersion implements GET method for /1.0/ovsdb/schema/<db-name>/expected. It returns
 // expected schema version of the specified database on this MicroOVN node. Expected schema version is
 // determined by looking at the schema file that is bundled with OVN/OVS packages.
 // URL variable <db-name> is expected to be one of the keys from supportedDBs.
 func getExpectedSchemaVersion(s *state.State, r *http.Request) response.Response {
-	requestedDB, err := url.PathUnescape(mux.Vars(r)["db"])
-	if err != nil {
-		logger.Errorf("failed to parse requested DB name from url '%s'", r.URL)
-		return response.ErrorResponse(500, "Internal Server Error")
-	}
-	requestedDB = strings.ToLower(requestedDB)
-
-	dbType, ok := supportedDBs[requestedDB]
-
-	if !ok {
-		return response.BadRequest(fmt.Errorf("DB '%s' not supported", requestedDB))
-	}
-
-	dbSpec, err := ovnCmd.NewOvsdbSpec(dbType)
-	if err != nil {
-		logger.Errorf("%s", err)
-		return response.ErrorResponse(500, "Internal Server Error")
+	dbSpec, errResponse := parseDbSpec(r)
+	if errResponse != nil {
+		return errResponse
 	}
 
 	expectedVersion, err := ovsdb.ExpectedOvsdbSchemaVersion(s, dbSpec)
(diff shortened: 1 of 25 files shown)


microovn/ovn/commands.go
```diff
@@ -16,12 +16,14 @@ const defaultDBConnectWait = 30 //Default time to wait for connection to ovsdb
...
-	Target string
-	Name   string
+// OvsdbSpec is a helper structure that encapsulates properties of an OVN/OVS database.
+type OvsdbSpec struct {
+	SocketPath   string // Path to a socket that the database has open
+	Name         string // Name of the database within the db file (e.g. "OVN_Northbound")
+	ShortName    string // Shorthand name for the database (e.g. "nb")
+	FriendlyName string // Human friendly name of the database ideal for logging purposes (e.g. "Northbound")
+	Schema       string // Path to a schema file for the database
+	IsCentral    bool   // Whether the database is used by OVN central services
```
fnordahl, line 26: Adding a bool for this purpose feels redundant since we already have a notion of OvsdbType.

Can we record the OvsdbType instead and perhaps call a helper function which knows which of the types are used for central DBs?

microovn/ovn/commands.go
```diff
@@ -16,12 +16,14 @@ const defaultDBConnectWait = 30 //Default time to wait for connection to ovsdb
...
 
-// ovsdbSpec is a helper structure for precise identification of ovsdb databases. A lot of
-// ovn/ovs commands take path to either database file, database socket or process control socket
-// along with the database name. This structure can be used for such cases.
-type ovsdbSpec struct {
-	Target string
-	Name   string
+// OvsdbSpec is a helper structure that encapsulates properties of an OVN/OVS database.
+type OvsdbSpec struct {
+	SocketPath   string // Path to a socket that the database has open
```
fnordahl, line 21: While I see that the code is currently hard coded for Unix sockets, it feels limiting knowing the underlying tooling accepts any type of target.

OTOH, if we will only ever use this for sockets i guess its fine and we can always change it later.

microovn/ovn/cli/commands.go
```diff
@@ -1,4 +1,4 @@
-package ovn
+package cli
```
fnordahl, line 1: The name `cli` is a bit misleading, is this package implementing a CLI?

If the goal is to shorten the `commands` name, would `cmd` work as well?

microovn/ovn/bootstrap.go
```diff
@@ -131,7 +132,7 @@ func Bootstrap(s *state.State, initConfig map[string]string) error {
 		return err
 	}
 
-	_, err = VSCtl(
+	_, err = cli.VSCtl(
```
fnordahl, line 135: The imported name is a bit ambiguous, should we perhaps use the golang alias functionality and import the package with a prefix so that it could read `ovn_cmd.VSCtl()`

fnordahl (commented):
Thanks for your work on this, Martin! Have unfortunately not been able to help test, but the approach looks good to me.

Do we need some additional CLI knobs to allow the human operator to inform themselves about the progress? For example showing the expected schema version in cluster status might help them spot whether a snap upgrade is missing on one of the nodes?

microovn/ovn/commands.go
mkalcok: I see what you mean and it's a good point. The reason I wanted to change it from the previous name `Target` was that some commands, like `ovn-appctl`, take "target" argument which points to `.ctl` file as opposed to `.sock`. I wanted to avoid confusions there.

I'll think of some better name, and I'm open to suggestions. Perhaps a `LocalSocketPath`?

If we have ever a need for using tcp/ssl "targets", we can extend this struct later.

mkalcok: Come to think about this, the `LocalSocketPath` wouldn't be the best option. If we ever need to use network targets for the database connection, we could probably add new `OvsdbType`, something like `OvsdbTypeNBCluster` and the constructor for `OvsdbSpec` would populate this field with correct tcp/ssl URLs. So perhaps this field could be called `SocketURL` and it would work for `unix:`, `tcp:` and `ssl:`

microovn/ovn/commands.go
mkalcok: I guess that this is a matter of perspective. If we replaced `IsCentral` with something like `DbType`, that would hold `OvsdbType`, then we would still have same number of members in the `OvsdbSpec`, but now we'd also need a helper method that would decide whether an instance of `OvsdbSpec` is central or not.

In a sense, we already have a helper like this. It's the constructor for the `OvsdbSpec`. You pass `OvsdbType` to `NewOvsdbSpec` and it fills in the `IsCentral` member with true/false.

microovn/ovn/cli/commands.go
mkalcok: Thank you, yes. I had my doubts about the `cli` as well. The `cmd` alternative (with `ovn_cmd` alias usage) is much better.

mkalcok:
> Do we need some additional CLI knobs to allow the human operator to inform themselves about the progress? For example showing the expected schema version in cluster status might help them spot whether a snap upgrade is missing on one of the nodes?

I like this idea. It would probably took another API endpoint. But we could add information about the schema upgrade status. I'll think about way to present this to the user in `microovn status` output, because just showing expected DB schema version might not cut it. It might be confusing to display information about expected NB schema version on a node that doesn't run "central" service.

microovn/ovn/commands.go
fnordahl: Yea, I guess it is a matter of preference of using methods that understand the data structure rather than accessing the data directly from foreign functions/method.

Right now it does not make much difference so I'm fine either way, but if we somewhere down the road need to make other decisions on the back of the different types we might find it makes more sense to store the type.

microovn/api/types/ovsdb.go
```diff
@@ -0,0 +1,118 @@
...
+	msg += fmt.Sprintf("Currently running schema: %s\n", s.RunningVersion)
+	msg += "Cluster report (expected schema versions):\n"
+	for _, node := range s.Nodes {
+		nodeName, err := addrToName(node.Host)
+		if err != nil {
+			nodeName = node.Host
+		}
+		msg += fmt.Sprintf("\t%s: ", nodeName)
+		if node.Error {
+			msg += "N/A\n"
```
fnordahl, line 111: This message is printed in multiple situations, and it does not appear to be particularly helpful for the operator.

1) For example when only one node of the cluster has been upgraded, the non-upgraded nodes will be shown as `N/A` regardless of them having central components that actually do need to be upgraded or not, example:
```
$ sudo microovn status
MicroOVN deployment summary:
- micro-1 (10.5.0.2)
  Services: central, chassis, switch

- micro-2 (10.5.0.3)
  Services: central, chassis, switch

- micro-3 (10.5.0.4)
  Services: central, chassis, switch

- micro-4 (10.5.0.5)
  Services: chassis, switch

OVN Database summary:
OVN Southbound: Upgrade or attention required!
Currently running schema: 20.21.0
Cluster report (expected schema versions):
    micro-1: 20.33.0
    micro-3: N/A
    micro-2: N/A
    micro-4: N/A

OVN Northbound: Upgrade or attention required!
Currently running schema: 6.1.0
Cluster report (expected schema versions):
    micro-1: 7.3.0
    micro-2: N/A
    micro-3: N/A
    micro-4: N/A
```

2)  When all central nodes are upgraded and the chassis nodes are not upgraded, the schema conversion will (rightfully) be held back, but the status does not reveal any information to the end user that this is the reason for the upgrade not happening.

```
$ sudo microovn status
MicroOVN deployment summary:
- micro-1 (10.5.0.2)
  Services: central, chassis, switch

- micro-2 (10.5.0.3)
  Services: central, chassis, switch

- micro-3 (10.5.0.4)
  Services: central, chassis, switch

- micro-4 (10.5.0.5)
  Services: chassis, switch

OVN Database summary:
OVN Southbound: Upgrade or attention required!
Currently running schema: 20.21.0
Cluster report (expected schema versions):
    micro-1: 20.33.0
    micro-4: N/A
    micro-3: 20.33.0
    micro-2: 20.33.0

OVN Northbound: Upgrade or attention required!
Currently running schema: 6.1.0
Cluster report (expected schema versions):
    micro-1: 7.3.0
    micro-4: N/A
    micro-3: 7.3.0
    micro-2: 7.3.0
```

So for the chassis node could we perhaps print a hint about snap upgrade being required? Another possiblitiy is to use the word "supported" instead of "expected" for schema version, that way it would not be confusing when we print schema version for the "client" nodes?

Mock-up:
```
Cluster report (supported schema versions):
    micro-1: 7.3.0
    micro-4: Unknown - snap upgrade required
    micro-3: 7.3.0
    micro-2: 7.3.0

Cluster report (supported schema versions):
    micro-1: 7.3.0
    micro-4: 6.1.0 - snap upgrade required
    micro-3: 7.3.0
    micro-2: 7.3.0
```

fnordahl (commented):
I took this for a spin, and apart from the comment in-line, the ergonomics of this when knowing how it works is good, great job on that! If we resolve the remaining issues this will be a fantastic UX imho.

Note that I did not do any thorough validation of data plane interruptions so I trust that to work as expected.

microovn/api/types/ovsdb.go
mkalcok: Thank you for the review and testing @fnordahl. For starters I want to clarify that every node, regardless of whether it runs central services or not, is expected to report "expected version" for the schema. However, we are facing kind of a one-off situation here where non-upgraded nodes do not have the API required for this feature to work. Therefore any non-upgraded nodes report `N/A`, in the future these nodes would report older schema version.
We can probably distinguish `404` response code and print more informative message in this scenario. That would leave the `N/A` value solely for unexpected errors during the expected schema version fetch.

So in the current scenario, where some hosts in the cluster do not support this feature you'd see something like:

OVN Database summary:
OVN Southbound: Upgrade or attention required!
Currently running schema: 20.21.0
Cluster report (expected schema versions):
    micro-1: 20.33.0
    micro-4: API not supported - MicroOVN requires upgrade
    micro-3: 20.33.0
    micro-2: 20.33.0

mkalcok:
I updated this PR based on the discussions we had above @fnordahl. Key points:

Distinguishing between unreachable or otherwise faulty cluster members and cluster members that lack API to report expected schema version:
```
OVN Database summary:
OVN Southbound: Upgrade or attention required!
Currently running schema: 20.21.0
Cluster report (expected schema versions):
	movn1: 20.33.0
	movn4: Error. Failed to contact member
	movn3: Missing API. MicroOVN needs upgrade
	movn2: Missing API. MicroOVN needs upgrade

OVN Northbound: Upgrade or attention required!
Currently running schema: 6.1.0
Cluster report (expected schema versions):
	movn1: 7.3.0
	movn4: Error. Failed to contact member
	movn2: Missing API. MicroOVN needs upgrade
	movn3: Missing API. MicroOVN needs upgrade
```

Added current schema version to "OK" output:
```
OVN Database summary:
OVN Southbound: OK (20.33.0)
OVN Northbound: OK (7.3.0)
```

Added functional tests to verify that schema upgrade gets triggered only after every host in the cluster is upgraded.

mkalcok:
Bit of a side-note: I talked to you offline @fnordahl about adding the DB schema upgrade test into the `upgrade.bats` and reusing containers, but I went back on that idea. It didn't really saved much time as the time-consuming part is installing/bootstrapping MicroOVN, not creating containers. Other drawbacks:
* harder parallelization in CI
* Messy setup/teardown that needs to clean up container after test

It wasn't a good idea and we should stick with single-use containers in the tests.

mkalcok:
Thank you @fnordahl. Yes I had it cherry-picked in this branch and forgot about it. I'll drop it and re-submit

mkalcok:
Two changes since last push:

1. Dropped the accidentally included commit 186c96f
2. Improved matching of HTTP response codes in `client.GetExpectedOvsdbSchemaVersion`

microovn/ovn/cmd/commands.go
```diff
@@ -200,18 +200,18 @@ func ControllerCtl(s *state.State, args ...string) (string, error) {
...
 // to the "ovsdb-client". Note that it is not necessary to pass "-t" argument, as the timeout is automatically included.
-func OvsdbClient(s *state.State, dbSpec *OvsdbSpec, connectTimeout int, resultTimeout int, args ...string) (string, error) {
-	err := WaitForDBState(s, dbSpec, OvsdbConnected, connectTimeout)
-	if err != nil {
-		return "", err
+func OvsdbClient(s *state.State, resultTimeout int, args ...string) (string, error) {
+	cert, key := paths.PkiClientCertFiles()
+	arguments := []string{
+		"-t", strconv.Itoa(resultTimeout),
+		"-c", cert,
```
fnordahl, line 211: Is this really needed?

The `ovsdb-client` tool is capable of communicating with the cluster through the Unix socket on one of the nodes with a DB member in the cluster.

fnordahl (changes requested):
Great work! I have now tested an upgrade as well and this process does indeed avoid situations where the end user upgrades nodes in the wrong order. The UX with overview of the entire cluster from `microovn status` is also wonderful.

I have some initial comments and stopped the review at a question on why you need the `ovdb-client` to speak with the SSL sockets, the upstream cluster upgrade code does all communication through a local unix socket on one of the DB nodes.

Thinking about the tests, if we perhaps put some of the code into helper functions, we might fold it into the regular upgrade test file, conditionally enabling the schema tests based on environment, similar to the IPv6 tests for example?

mkalcok:
Thanks for the review @fnordahl 

> I have some initial comments and stopped the review at a question on why you need the ovdb-client to speak with the SSL sockets, the upstream cluster upgrade code does all communication through a local unix socket on one of the DB nodes.

main reason is to "make stuff work" on any cluster member, not just those that are running "central" service. Whether it's in the tests or in the `microovn status`, detection of currently running schema would not work on "switch-only" nodes if it tried to use local unix socket.

> Thinking about the tests, if we perhaps put some of the code into helper functions, we might fold it into the regular upgrade test file, conditionally enabling the schema tests based on environment, similar to the IPv6 tests for example?

I'll look into this again, but I'm afraid that it would greatly increase the test complexity. Regular upgrade test is `deploy -> upgrade all nodes -> test if stuff still works`. With these tests in there, we add bunch of other conditions and tests after each step of upgraded cluster member. I'm afraid it could make untangling test failures more difficult. But you do have a point, perhaps there's a way to save some time in this test.

microovn/ovn/cmd/commands.go
mkalcok: This addition makes this wrapper usable also on nodes that don't run central services.

tests/test_helper/bats/ovsdb_schema_upgrade.bats
```diff
@@ -0,0 +1,150 @@
...
+        current_nb=$(get_current_ovsdb_schema_version "$container" "nb")
+        current_sb=$(get_current_ovsdb_schema_version "$container" "sb")
+        assert [ "$current_nb" == "$original_nb" ]
+        assert [ "$current_sb" == "$original_sb" ]
+    done
+
+    # After all cluster members are upgraded, verify that the cluster reports upgraded schema version
+    # as well.
+    local timeout=30
+    local err=1
```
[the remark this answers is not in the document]
mkalcok: 👍  I guess I wasn't sure about exact rules vor variable scoping in bash so I went with safer approach :D

microovn/ovn/cmd/commands.go
fnordahl: That is not a good reason. The test code already defines lists with central nodes and chassis nodes, so you know where you can get the information. The test also defines a "probe container" which also has this information locally.

fnordahl:
> Thanks for the review @fnordahl
> 
> > I have some initial comments and stopped the review at a question on why you need the ovdb-client to speak with the SSL sockets, the upstream cluster upgrade code does all communication through a local unix socket on one of the DB nodes.
> 
> main reason is to "make stuff work" on any cluster member, not just those that are running "central" service. Whether it's in the tests or in the `microovn status`, detection of currently running schema would not work on "switch-only" nodes if it tried to use local unix socket.

You still have the "just works" capability by using the REST API to call to one of the nodes with the databases?

> > Thinking about the tests, if we perhaps put some of the code into helper functions, we might fold it into the regular upgrade test file, conditionally enabling the schema tests based on environment, similar to the IPv6 tests for example?
> 
> I'll look into this again, but I'm afraid that it would greatly increase the test complexity. Regular upgrade test is `deploy -> upgrade all nodes -> test if stuff still works`. With these tests in there, we add bunch of other conditions and tests after each step of upgraded cluster member. I'm afraid it could make untangling test failures more difficult. But you do have a point, perhaps there's a way to save some time in this test.

There are indeed good arguments to keep them separate too, just wanted to explore the possibilities of code reuse by putting blocks of code that belong together into functions instead of in the body of the test definition.

The foundation of the two tests are very similar. We are in the process of adding a third upgrade test which needs to use the same foundation and just add on data path validation.

tests/test_helper/bats/ovsdb_schema_upgrade.bats
```diff
@@ -0,0 +1,150 @@
...
+    local old_member_msg_sb=""
+    if [ "$MICROOVN_SNAP_REV" -le 395 ]; then
+        old_member_msg_nb="Missing API. MicroOVN needs upgrade"
+        old_member_msg_sb="Missing API. MicroOVN needs upgrade"
+    else
+        old_member_msg_nb="$original_nb"
+        old_member_msg_sb="$original_sb"
+    fi
+
+    for container in $TEST_CONTAINERS; do
```
fnordahl, line 63: This loop iterates over all the containers and moves in a forward direction. This means that the current element is to be upgraded, any previous elements are already upgraded and any coming elements are not upgraded yet. As such we can infer all this information from the current array index.

Would it make sense to use that instead of keeping track of all these extra arrays? Something like:
```bash
read -r -a containers <<< "$TEST_CONTAINERS"
for ((i = 0; i < "${#containers[@]}"; i++)) {
    container=${containers[i]}
    echo "processing container $container"

    for ((j = 0; j < i; j++)) {
        echo "upgraded container ${containers[j]}"
    }

    for ((j = i + 1; j < "${#containers[@]}"; j++)) {
        echo "old container ${containers[j]}"
    }
    echo
}
```

fnordahl (changes requested):
To move forward I'd like to move the responsibility of retrieving information about the cluster from the server side API implementation to the client implementation.

In addition to avoiding cascading/scaling issues on the server side it would enable the client to direct queries about the running schema version to the nodes that has this information, removing the need for using credentials and calling over the network with `ovsdb-client`.

On the use of credentials, while your proposal works, we need to consider the consequences of using them in new places. What stands out as problematic to me is the expectation that chassis nodes are allowed privileged access to the NB DB.

The NB API is privileged, and it is intended that only the CMS has access to it. The data path components are expected to only speak to the SB DB.

While there are no restrictions imposed on this today, we need to be able to support stricter requirements around this in the future.

tests/test_helper/bats/ovsdb_schema_upgrade.bats
mkalcok: This looks reasonable +1

microovn/api/ovsdb/schema_version.go
```diff
@@ -1,64 +1,174 @@
...
-	if !ok {
-		return response.BadRequest(fmt.Errorf("DB '%s' not supported", requestedDB))
+// getSchemaVersionStatus implements GET method for /1.0/ovsdb/schema/<db-name>/status. It returns information about
+// currently used OVSDB schema version and a list of schema version that each cluster member is expecting to run.
+//
+// This output can be used to determine if there are any discrepancies between the expected schema versions on each
+// cluster member and the version that's actually used. If some members of the cluster expect lower schema version
+// than their peers, it might be a sign that they are running older version of MicroOVN (and by extension, OVN) and
+// need to be upgraded.
+func getSchemaVersionStatus(s *state.State, r *http.Request) response.Response {
```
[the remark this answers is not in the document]
mkalcok: Core of this implementation can definitely be moved to the client side. Though I'm not entirely clear on what we'd be achieving in terms of scaling. Current implementation results in 1 call from client to server and then `($cluster_size - 1)` calls from the server to other servers in the deployment.
Moving the implementation to client would result in `$cluster_size` number of calls from client to servers. So at the end, there'd be the same number of calls.

microovn/api/types/ovsdb.go
```diff
@@ -0,0 +1,128 @@
...
+
+// UpgradeRequired method returns "true" if there's any discrepancy between currently running OVSDB schema version
+// and a schema version expected by the individual hosts in the cluster. It also returns "true", if any of the cluster
+// members had Error in their OvsdbSchemaNodeStatus as it may indicate either a host that needs an attention because
+// it's down, or it's an older version of the MicroOVN that does not support API for fetching expected schema version.
+//
+// If all values match and there was no error, this method returns "false".
+//
+// This method returns error if Nodes is an empty list.
+func (s *OvsdbSchemaClusterStatus) UpgradeRequired(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
mkalcok: This is a method implemented directly on the API result type `OvsdbSchemaClusterStatus`. I think it makes more sense to implement is as method, rather than a free function somewhere else that takes `OvsdbSchemaClusterStatus` as an argument. I'm open to suggestions on alternative implementation, thought it may not matter that much if we end up re-implementing `getSchemaVersionStatus` API, this type may go away entirely.

microovn/api/types/ovsdb.go
```diff
@@ -0,0 +1,128 @@
...
+	return err != nil, nil
+
+}
+
+// Format returns a string that verbosely describes contents of the OvsdbSchemaClusterStatus.
+//
+// If the cluster does not require update, a simple "OK" is returned. Otherwise, this method returns string
+// that describes currently running OVSDB schema version, schema versions expected by the individual cluster members,
+// and whether there were any errors when fetching the expected schema from other cluster members.
+func (s *OvsdbSchemaClusterStatus) Format(ctx context.Context, cli *client.Client) string {
```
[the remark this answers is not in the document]
mkalcok: I was thinking about naming it `String`, but again this type may end up going away.

mkalcok:
Thank you for the feedback

>On the use of credentials, while your proposal works, we need to consider the consequences of using them in new places. What stands out as problematic to me is the expectation that chassis nodes are allowed privileged access to the NB DB.
>
>The NB API is privileged, and it is intended that only the CMS has access to it. The data path components are expected to only speak to the SB DB.
>
>While there are no restrictions imposed on this today, we need to be able to support stricter requirements around this in the future.

While I didn't fully understand/agree with the argument about the scaling issue of the current implementation. This other argument is imo good enough on its own, to warrant creating MicroOVN API for "currently running schema version". I'll get on it.

mkalcok:
Relevant changes since last review:

* MicroOVN no longer uses ssl endpoints when calling `ovsdb-client` wrapper
  * This was needed to get active Southbound/Northbound schema versions, but it was replaced with MicroOVN API endpoint that now provides this value
* API endpoint `getOvsdbClusterStatus` was replaced with `getAllExpectedSchemaVersions` which now focuses only on getting the expected schema version from deployment nodes. It no longer includes active schema version in its reply.
  * Similarly, its API return type was simplified and now only returns list of nodes and their expected schema versions
* Logic around interpreting API return types was moved from `types` package and now lives in the `cmd.microovn` package

One thing that you raised before @fnordahl, was that when request needs to be made to all nodes, client [sends request only to one server](https://github.com/mkalcok/microovn/blob/2f55637f7a5df008036a1d4bbd729473802fa465/microovn/client/client.go#L98), and that server then [fans out requests to the rest of the nodes, collects results, and returns them to original client](https://github.com/mkalcok/microovn/blob/2f55637f7a5df008036a1d4bbd729473802fa465/microovn/api/ovsdb/schema_version.go#L101-L112). That approach was kept in this iteration as well because the MicroOVN's client/command packages (`microovn.client`/`microovn.cmd.microovn`)  are very lightweight. They don't use objects like `state.State` which allows, for example, [to make direct DB queries](https://github.com/mkalcok/microovn/blob/2f55637f7a5df008036a1d4bbd729473802fa465/microovn/node/services.go#L71), they rely purely on communicating via the http API.
This prevents them from easily fetching info about nodes in the cluster and their services. They also can't use very convenient [`State.Cluster()` function that returns client instances for every node in the cluster](https://github.com/mkalcok/microovn/blob/2f55637f7a5df008036a1d4bbd729473802fa465/microovn/api/ovsdb/schema_version.go#L94).
If we really wanted, we could implement APIs for every information about cluster we need. Then in `client`/`cmd` package we could query that API, fetch required information and build client instances manually, but I think that's just unnecessary work.

I also think that it's overall cleaner design to keep code that understands internals on the server side, rather than to have that logic on the client side.

microovn/ovn/ovsdb/ovsdb.go
```diff
@@ -26,3 +41,241 @@ func ExpectedOvsdbSchemaVersion(s *state.State, dbSpec *ovnCmd.OvsdbSpec) (strin
...
+
+	if !centralActive {
+		logger.Infof(
+			"OVN %s database won't be upgraded from this node because 'central' service is not active",
+			dbSpec.FriendlyName,
+		)
+		return nil
+	}
+
+	backOffMs := 100
```
fnordahl, line 207: nit: The number used for back off could be a constant, which would allow the other checks used for time to be multipliers of the constant. If we need to change it, we can change it in one place and the rest of the algorithm will adjust in relation to the constant.

microovn/ovn/ovsdb/ovsdb.go
```diff
@@ -26,3 +41,241 @@ func ExpectedOvsdbSchemaVersion(s *state.State, dbSpec *ovnCmd.OvsdbSpec) (strin
...
+				logger.Infof(
+					"This host is not a designated to upgrade OVN %s DB schema. Rechecking schema status in %d",
+					dbSpec.FriendlyName,
+					backOffMs,
+				)
+			}
+		}
+
+		time.Sleep(time.Duration(backOffMs) * time.Millisecond)
+		if backOffMs < 2000 {
```
fnordahl, line 278: nit: see comment above about using constant and multipliers instead of bare numbers for this algorithm.

microovn/ovn/ovsdb/ovsdb.go
```diff
@@ -26,3 +41,241 @@ func ExpectedOvsdbSchemaVersion(s *state.State, dbSpec *ovnCmd.OvsdbSpec) (strin
...
+			backOffMs *= 2
+		}
+
+		if backOffMs > 2000 {
+			backOffMs = 2000
```
fnordahl, line 283: nit: see comment above about using constant and multipliers instead of bare numbers for this algorithm.

fnordahl (commented):
Thanks for updating the series!

Not introducing a new place for using credentials and keeping the cluster schema operations local makes it more palatable to me.

On keeping the fan-out behavior in the daemon, it's not my preferred approach and we might get into trouble if there are communication issues to some nodes and/or if users do something silly on large scale clusters.

But to be fair, the underlying plumbing does make use of goroutines for handling requests so we at least won't block, and the API endpoints are relatively light weight (keeping the DB cluster communications local helped there too).

If you deem the cost of moving this to the client to be too high what we have now is an acceptable compromise for me.

mkalcok:
I went to to bottom addressing your suggestions @fnordahl and I believe I got them all.

tests/test_helper/ovsdb.bash
```diff
@@ -0,0 +1,81 @@
...
+            connection_string="unix:/var/snap/microovn/common/run/ovn/ovnsb_db.sock"
+        else
+            echo "# Unknown database type '$nbsb'. Valid values: 'nb', 'sb'"
+            return 1
+    fi
+
+    local pki_dir="/var/snap/microovn/common/data/pki/"
+    local cert="$pki_dir/client-cert.pem"
+    local key="$pki_dir/client-privkey.pem"
+    local ca_cert="$pki_dir/cacert.pem"
```
fnordahl, line 31: nit: these are no longer needed

fnordahl (approved):
Thank you for making the upgrade experience of MicroOVN great!

PR #132, opened 2024-05-29
tests: Use alias for containers' release version

mkalcok:
Instead of using hardcoded name of the Ubuntu release that's used for test containers, we can leverage alias "lts" that will keep pointing to current LTS release.

-----
Alternatively, we can use `ubuntu:default` that will point to the latest release, but we went with `jammy` so far and I don't think it matters too much for the tests.

### DIFF
--- a/tests/test_helper/lxd.bash
+++ b/tests/test_helper/lxd.bash
@@ -1,10 +1,9 @@
 function launch_containers() {
-    local image_name=$1; shift
     local containers=$*
-
+    local image="${MICROOVN_TEST_CONTAINER_IMAGE:-ubuntu:lts}"
     for container in $containers; do
-        echo "# Launching $container" >&3
-        lxc launch -q "ubuntu:$image_name" "$container" < \
+        echo "# Launching '$image' container: $container" >&3
+        lxc launch -q "$image" "$container" < \
             "$BATS_TEST_DIRNAME/lxd-instance-config.yaml"
     done
 }
--- a/docs/developers/testing.rst
+++ b/docs/developers/testing.rst
@@ -77,6 +77,17 @@ To run individual test suites you can execute:
 
    make tests/<name_of_the_test_suite>.bats
 
+By default, functional tests run in LXD containers based on ``ubuntu:lts``
+image. This can be changed by exporting environment variable
+``MICROOVN_TEST_CONTAINER_IMAGE`` and setting it to a valid LXD image name.
+
+For example:
+
+.. code-block:: none
+
+    export MICROOVN_TEST_CONTAINER_IMAGE="ubuntu:jammy"
+    make check-system
+
 .. tip::
 
    If your hardware can handle it, you can run test suites in parallel by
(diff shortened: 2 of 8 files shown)


tests/test_helper/lxd.bash
```diff
@@ -1,10 +1,9 @@
 function launch_containers() {
-    local image_name=$1; shift
```
fnordahl, line 2: I see no reason to remove this argument, we might need tests targeting explicit versions?

tests/test_helper/lxd.bash
```diff
@@ -1,10 +1,9 @@
...
     local containers=$*
 
     for container in $containers; do
         echo "# Launching $container" >&3
-        lxc launch -q "ubuntu:$image_name" "$container" < \
+        lxc launch -q "ubuntu:lts" "$container" < \
```
fnordahl, line 6: The `lts` alias appears to be a reasonable default for the main branch, however we might want to lock down stable branches to a specific version.

What about retrieving the default from an environment variable, which defaults to `lts` when not set?

fnordahl (changes requested):

tests/test_helper/lxd.bash
mkalcok: I wanted to cut down on moving parts. I was thinking about possible scenarios, and came to the conclusion that the choice of the image version would likely be global for all tests in the suite, rather than a configurable option that every test sets individually.
That being said, I'm not dead set on removing it.

tests/test_helper/lxd.bash
mkalcok: Sure, env variable could be one way to set a default value. Thought, do we really have a use-case for configurable test images? The way I see it is that microovn, being snap, is heavily insulated from the underlying OS version so it makes very little difference what container version we use for testing. The only possible thing that could make a difference is a kernel version, but that can't be affected by specifying different container image. We'll still share the underlying host's kernel, that we can't influence.

My main intention with this change was to reduce the amount of stuff we need to think about, and change, whenever there's a new release.

fnordahl: We do indeed get pretty good insulation from the environment being a confined snap, there are however other factors that come into play such as apparmor and difference in system defaults (found a [random bug](https://bugs.launchpad.net/snapd/+bug/1981546) as an example).

Not having to change things on new releases is a high priority, but we also don't want stable branches to break.

I'm fine with trying to have this rolling, but I'd really like to have the ability to control this via the environment from the start, so that we don't need to backport that change once a stable branch breaks.

tests/test_helper/lxd.bash
fnordahl: I guess that's fair, less is more, we can add it back if we actually need individual control.

tests/test_helper/lxd.bash
mkalcok: That is a very good point, and I was thinking about running regular (weekly?) CI runs for all branches, to discover these issue as they happen, not just when we have some backports. We could then run all stable branches with all available LTSes on regular bases.

fnordahl (approved):
LGTM, thanks!

PR #133, opened 2024-05-29
CI: Run lightweight tasks on Github-hosted runners

mkalcok:
We switched all our jobs to self-hosted runners when we experienced very long queues on GH-hosted runners. However, there should now be enough GH-hosted runners for us to safely run the lightweight tasks on them again.

### DIFF
--- a/.github/workflows/commits.yml
+++ b/.github/workflows/commits.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   cla-check:
-    runs-on: [self-hosted, linux, X64, jammy, large]
+    runs-on: ubuntu-latest
     permissions:
       pull-requests: read
     name: Canonical CLA signed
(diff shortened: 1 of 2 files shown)


fnordahl (commented):
Can you elaborate a bit on the motivation for this change?

Looking at recent runs the execution time appears to be pretty good as-is?

AFAIR the issue was that our GitHub org was out over quota and that consuming idle private cloud resources through self-hosted runners is preferred over increasing the quota.

If the org is currently within quota, the GitHub hosted runners would indeed be more light weight and probably quicker, but would that not just be a side effect of most jobs being moved to self-hosted runners? And what would happen if we start moving back?

fnordahl (approved):
LGTM, thanks!

With reference to my previous comment the self-hosted runners apparently got too popular, so we have indeed been asked to offload lighter jobs to the cloud. (I guess I should read memoes sent during vacation before doing reviews ;)

PR #135, opened 2024-06-03
ovs wrappers: Inject SSL arguments.

mkalcok:
This change adds a common wrapper for ovs commands that support client connection over ssl:
  * ovs-ofctl
  * ovs-vsctl
  * ovsdb-client

This wrapper automatically inject arguments for client certificate, private key and CA certificate. It eliminates the need for user to type them in every time and does not cause any errors/warnings if the commands use other connection methods than SSL.


fnordahl (approved):
LGTM, thanks!

Minor nit: if someone wants to have a look at the output of `--help` as an unprivileged user, this will generate error messages that were previously not there.

mkalcok:
That's a good point! thank you. I think we can work around that in the wrapper by checking if cert files are readable

snapcraft/commands/ovs-ssl-wrapper
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+export OVS_RUNDIR="${SNAP_COMMON}/run/switch/"
+
+# Optionally load SSL certificates if they are readable
+if [ -r "$SNAP_COMMON/data" ]; then
```
fnordahl, line 5: In an aged install, the `$SNAP_COMMON/data` directory is readable while `${OVN_PKI_DIR}` is not. So in my example we would still see this error.

Would it be more appropriate to check whether the files within `${OVN_PKI_DIR}` are accessible?

Looking at `${SNAP}/ovn.env` it sources `"${SNAP_COMMON}/data/env/ovn.env"` which is probably what lead you to do the check on `"$SNAP_COMMON/data"` but with the inconsistency we have between fresh and older deployments I don't think we can.

Sourcing `/snap/microovn/current/ovn.env` does populate `OVN_PKI_DIR` even if it will complain about not being able to read `"${SNAP_COMMON}/data/env/ovn.env"`, so we could do that and deal with permissions for the data dir and/or permission check for `/snap/microovn/current/ovn.env` separately, whichever we find more appropriate.

mkalcok: > Looking at ${SNAP}/ovn.env it sources "${SNAP_COMMON}/data/env/ovn.env" which is probably what lead you to do the check on "$SNAP_COMMON/data"

Exactly.

> Sourcing /snap/microovn/current/ovn.env does populate OVN_PKI_DIR even if it will complain about not being able to read

We could also conditionally source the `$SNAP_COMMON/data/env/ovn.env` in `$SNAP/ovn.env`. No sense in sourcing something that can't be read. We could then properly gate the `ovs-ssl-wrapper` based on accesibility to `$OVN_PKI_DIR`

snapcraft/ovn.env
```diff
@@ -1,7 +1,10 @@
 # Set of environment variables for OVN commands
 
-# Load runtime environment variables
-. "${SNAP_COMMON}/data/env/ovn.env"
+# Load runtime environment variables if available
+runtime_env="${SNAP_COMMON}/data/env/ovn.env"
+if [ -r "$runtime_env" ]; then
+    . "${SNAP_COMMON}/data/env/ovn.env"
```
fnordahl, line 6: There is a variable containing this string above, is not using it here intentional?

mkalcok: Thank you, I overlooked it

fnordahl (approved):
LGTM, thanks!

PR #138, opened 2024-06-11
Measure data path uptime as part of upgrade test.

fnordahl:

### DIFF
--- a/tests/test_helper/setup_teardown/upgrade.bash
+++ b/tests/test_helper/setup_teardown/upgrade.bash
@@ -38,6 +38,13 @@ setup_file() {
         fi
     done
 
+    # Make sure that microcluster is fully converged before proceeding.
+    # Performing further actions before the microcluster is ready may lead to
+    # unexpectedly long convergence after a microcluster schema upgrade.
+    for container in $TEST_CONTAINERS; do
+        wait_microovn_online "$container" 60
+    done
+
     # detect and export initial MicroOVN snap revision
     container=$(echo "$TEST_CONTAINERS" | awk '{print $1;}' )
     export MICROOVN_SNAP_REV=""
@@ -61,6 +68,8 @@ setup_file() {
         echo "# Upgrading MicroOVN from revision $MICROOVN_SNAP_REV " \
              "on chassis container(s)." >&3
         install_microovn "$MICROOVN_SNAP_PATH" $CHASSIS_CONTAINERS
+
+        wait_microovn_online "$container" 60
     fi
 }
 
--- a/tests/test_helper/setup_teardown/upgrade.bash
+++ b/tests/test_helper/setup_teardown/upgrade.bash
@@ -48,12 +48,7 @@ setup_file() {
         echo "# Upgrading MicroOVN from revision $MICROOVN_SNAP_REV" >&3
         install_microovn "$MICROOVN_SNAP_PATH" $TEST_CONTAINERS
 
-        for container in $TEST_CONTAINERS; do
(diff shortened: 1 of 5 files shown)


tests/test_helper/bats/upgrade_dp.bats
```diff
@@ -0,0 +1,54 @@
...
+    # Remove namespace used for data path test
+    for container in $CHASSIS_CONTAINERS; do
+        microovn_delete_vif "$container" upgrade_dp0 upgrade_vif0
+        netns_delete "$container" upgrade_dp0
+        microovn_delete_gw_router "$container"
+    done
+}
+
+
+@test "MicroOVN control of schema conversion prevents data path downtime" {
```
mkalcok, line 39: Just a general thought here. We probably want to run this DP test on any upgrade, not just those that involve OVN schema upgrades. I'd say it's even more critical to ensure that in-channel snap upgrades don't cause DP downtime.
Otherwise this LGTM so far.

fnordahl: Thanks for feedback.

On in-channel upgrade tests: That would be nice, the upgrade tests are currently hard wired for cross channel checks though:
https://github.com/canonical/microovn/blob/833634c6a724d177a01ddef2a2957427ef26ce63/tests/test_helper/setup_teardown/upgrade.bash#L13

tests/test_helper/setup_teardown/upgrade.bash
```diff
@@ -38,30 +39,112 @@ setup_file() {
...
+            microovn_add_gw_router "$container"
+            netns_add "$container" "$UPGRADE_NS_NAME"
+            microovn_add_vif "$container" \
+                "$UPGRADE_NS_NAME" "$UPGRADE_VIF_NAME"
+            ping_start "$container" 10.42.${ctn_n}.1 "$UPGRADE_NS_NAME"
+        done
+
+        echo "# Upgrading MicroOVN from revision $MICROOVN_SNAP_REV "\
+             "on central container(s)." >&3
+        install_microovn "$MICROOVN_SNAP_PATH" $CENTRAL_CONTAINERS
```
mkalcok, line 77: Upgrading central containers in the `file_setup` may be unexpected for the tests, and in fact breaks the "ovsdb schema upgrade" checks in a very subtle way. The schema check relies on collecting schema version before upgrade
https://github.com/canonical/microovn/blob/873f059582140da4c187ba347cb7f231c3aa69f3/tests/test_helper/bats/ovsdb_schema_upgrade.bats#L33-L45

If the snap is already updated by the time the test gets to it, it will always skip the rest of the check. I wonder if it wouldn't be better to make "datapath uptime" checks into it's own test suite.

mkalcok (changes requested):
Thanks for these tests @fnordahl, the'll be great for ensuring that our upgrades are not causing datapath outages. However there's an issue with how they are organized currently. Please see my inline comment

tests/test_helper/setup_teardown/upgrade.bash
fnordahl: The `file_setup` upgrade is only performed when `UPGRADE_DO_UPGRADE` is set. This is not the case for the ovsdb schema upgrade test. So how would it break it?

mkalcok: Oh, I see, my bad. I didn't notice the `UPGRADE_DO_UPGRADE` usage.

mkalcok (approved):
LGTM, thanks!

PR #139, opened 2024-06-12
ci: Use GH-hosted runners for all jobs

mkalcok:
Since we, as an organisation, have access to more Github-hosted runners now, we've been encouraged to use them in favor of self-hosted runners.

Our tests are not particularly memory-hungry, so they should run in the GH-hosted runners comfortably.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
     name: build
     needs:
       - lint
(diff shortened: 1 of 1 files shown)


mkalcok:
Tests pass in on the GH hosted runners, and they even finish a bit faster than on the self-hosted. Additional benefit is that we won't have to manually approve test runs for external contributors. Overall I'd say this is a good change

fnordahl (approved):
LGTM, thanks!

PR #140, opened 2024-06-13
Make use of dynamic test registration

fnordahl:

### DIFF
--- a/tests/test_helper/bats/upgrade.bats
+++ b/tests/test_helper/bats/upgrade.bats
@@ -1,51 +1,10 @@
 # This is a bash shell fragment -*- bash -*-
 
+# Instruct the ``setup_file`` function to perform the upgrade.
+export UPGRADE_DO_UPGRADE=1
 load "test_helper/setup_teardown/$(basename "${BATS_TEST_FILENAME//.bats/.bash}")"
 
-setup() {
-    load test_helper/lxd.bash
-    load test_helper/common.bash
-    load test_helper/microovn.bash
-    load test_helper/upgrade_procedures.bash
-    load ../.bats/bats-support/load.bash
-    load ../.bats/bats-assert/load.bash
-
-    # Ensure TEST_CONTAINERS is populated, otherwise the tests below will
-    # provide false positive results.
-    assert [ -n "$TEST_CONTAINERS" ]
-}
-
-
-@test "Verify that currently released snap can be upgraded" {
-    echo "# Upgrading MicroOVN from revision $MICROOVN_SNAP_REV" >&3
-    install_microovn "$MICROOVN_SNAP_PATH" $TEST_CONTAINERS
-
-    for container in $TEST_CONTAINERS; do
-        local container_services
-        container_services=$(microovn_get_cluster_services "$container")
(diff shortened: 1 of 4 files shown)


tests/test_helper/bats/upgrade.bats
```diff
@@ -1,51 +1,10 @@
...
-    echo "# $output" >&3
-    echo "#" >&3
-    assert_success
-
-    run bats -F junit $tls_test_filename
-
-    echo "# $output" >&3
-    echo "#" >&3
-    assert_success
-}
```
mkalcok, line 51: I'm just wondering here whether it wouldn't be more readable to keep the `Verify that currently released snap can be upgraded` test and just replace the `run bats -F junit ...` hack with `load test_helper/bats/...`. Are there any benefits of removing this test by moving its "body" to the `setup_file` function?

One other potential drawback would also be less explicit output when running the tests. Running `upgrade.bats` would look like just like running `cluster.bats` and `tls_cluster.bats` combined.

mkalcok (commented):
Can this be replaced with dynamic loading as well?
https://github.com/fnordahl/microovn/blob/6ec00ac2a9710fc05dcce3d8073b05c69d86b37c/tests/scaleup_cluster.bats#L67-L75

Or would that just queue the tests at the end and we need, in this instance, run these tests "in-line", during every loop execution?

tests/test_helper/bats/upgrade.bats
fnordahl: Loading definitions inside a test is unfortunately not a thing.

The setup statements are asserted just like everything else and there is a comment at the very top pointing at where the upgrade steps are happening.

mkalcok: > Loading definitions inside a test is unfortunately not a thing.

I wasn't aware of this limitation...then all my questions are moot

mkalcok (approved):
I really like this change @fnordahl. Untangling the `junit` outputs from the "included" tests was not a pleasant experience. LGTM!

fnordahl:
> Can this be replaced with dynamic loading as well? https://github.com/fnordahl/microovn/blob/6ec00ac2a9710fc05dcce3d8073b05c69d86b37c/tests/scaleup_cluster.bats#L67-L75

Possibly, I looked at it briefly and it has a different mode than the other tests, so I'd save it for a future PR.

> Or would that just queue the tests at the end and we need, in this instance, run these tests "in-line", during every loop execution?

What we potentially can do is use the parametrization and pass a expected number of nodes to the tests, they can then conditionally all call into a `scale_to_n_nodes` helper. The iteration with arg 1 would then do nothing, the first test in the iteration with arg 2 scales the cluster and so on. But, let's save for a future PR.

mkalcok:
> What we potentially can do is use the parametrization and pass a expected number of nodes to the tests, they can then conditionally all call into a scale_to_n_nodes helper. The iteration with arg 1 would then do nothing, the first test in the iteration with arg 2 scales the cluster and so on. But, let's save for a future PR.

I was thinking about exactly the same approach right now. But as you said, that can be done in the future PRs.

PR #141, opened 2024-06-17
Prepare for 24.03

fnordahl:


fnordahl:
Marking this ready for review @mkalcok, but as we discussed oob, we probably want to merge an update to test matrix so that we test upgrades both from 22.03 and 24.03 before merging.

mkalcok (dismissed):
LGMT, thanks. One thing I was thinking about was to create short page in the `For Developers` section that would outline steps needed to take when cutting the stable branch and creating a new stable channel in snap store.
It can be either part of this PR, or you can merge this and we'll add the doc page separately.

mkalcok:
Re: test matrix.

I'll have the PR up shortly, we can discuss the approach there, but you are right, let's not merge this before we do the test change.

fnordahl:
> LGMT, thanks. One thing I was thinking about was to create short page in the `For Developers` section that would outline steps needed to take when cutting the stable branch and creating a new stable channel in snap store. It can be either part of this PR, or you can merge this and we'll add the doc page separately.

That's a good idea, I'll add it to this PR.

mkalcok (approved):
LGTM, thanks.
And sorry about the missing newline at the end of the file. My editor doesn't seem to care (or highlight those) so if I don't notice it in the PR diff, it gets in.

PR #143, opened 2024-06-25
Update microcluster dependency

masnax:
This attempts to address the concerns in #121 when adding the new api extensions feature from microcluster.

In particular, this fixes 2 bugs:

* The remaining cluster will not become unresponsive with a schema error while some nodes are awaiting a schema upgrade.
* If the database is inaccessible for any reason, it will report a `503` HTTP error.
  * If the database is in the process of starting, it will report `Database is still starting`.
  * If the database is not initialized, it will report `Database is not yet initialized`.
  * If the database is shut down, it will report `Database is offline`.
  * If the database is awaiting an upgrade, it will report `Database is waiting for an upgrade: %d cluster members have not yet received the update`.

Additionally, [GetUpgradingClusterMembers](https://github.com/canonical/microcluster/blob/f79afecf8b8945cd4500d3d0546e52671cc79aac/cluster/cluster_members.go#L277) can be used in the midst of an upgrade to fetch all existing cluster members, as well as a map of which cluster members are currently awaiting an upgrade.


mkalcok:
Hi @masnax. Thank you for this change. I think that it will greatly improve UX during upgrades. However I'm seeing some issues with the microovn daemon on upgrade.

Steps to reproduce:
* Setup 3 VMs/containers
* Install `microovn` from `24.03/beta` channel (to avoid any potential issues coming from OVN schema upgrade mixing in here)
* bootsrap/add/join to setup microovn cluster
* gradually upgrade each VM/container with microovn from your build

What I'm seeing is that shortly after you upgrade first host, the messages are displayed as you described them. However after some time (~30 seconds), the `microovn.daemon` service fails on the upgraded host. Logs show:
```
Jun 26 08:36:43 movn3 systemd[1]: snap.microovn.daemon.service: Failed with result 'exit-code'.

# then service tries to restart

Jun 26 08:36:43 movn3 microovn.daemon[8370]: time="2024-06-26T08:36:43Z" level=info msg="Daemon stopped"
Jun 26 08:36:43 movn3 microovn.daemon[8370]: Error: Daemon stopped with error: Control socket already present ("/var/snap/microovn/common/state/control.socket"); is another daemon already running?
Jun 26 08:36:43 movn3 systemd[1]: snap.microovn.daemon.service: Main process exited, code=exited, status=1/FAILURE

# cycle of restart attempts continues with above message until the threshold is reached
```

Even after upgrading all members and manually attempting to start `microovn.daemon` service, it fails with the same errors.

masnax:
OK, looks like the https://github.com/canonical/microcluster/pull/161 solved the upgrade failure. 

But now it looks like the `ovsdb_schema_upgrade` test is failing because it's waiting for all cluster members to report "ONLINE" after refreshing its snap, but since the other systems haven't refreshed yet, it fails because
```
Error: Database is waiting for an upgrade: 3 cluster members have not yet received the update
```

So I think we might need some branching validation here, since there may also be ovsdb schema upgrades that don't coincide with dqlite schema upgrades.

https://github.com/canonical/microcluster/pull/162 will add the current upgrade status of the cluster to the `Status` field reported by `microovn cluster list`, so I think we can check if `wait_microovn_online` reports this instead of `ONLINE`, and if so require all cluster members to upgrade before proceeding to verify the output of `microovn status`, as we can't yet be sure that all services are up.

mkalcok:
Thank you for digging into this @masnax. I suspected that the schema upgrade tests might be a problem, as they relied heavily on the previous behavior. Let me know if I can somehow help with tweaking of the tests or the ovsdb upgrade feature

masnax:
I've updated the ovsdb schema test to skip checking for OVN services if it encounters a dqlite schema update, since those services will not have started yet. 

`microovn cluster list` will now report the dqlite schema upgrade status:

```
+--------+----------------+-------+-----------------------------+---------------+
|  NAME  |    ADDRESS     |  ROLE |         FINGERPRINT         |    STATUS     |
+--------+----------------+-------+-----------------------------+---------------+
| micro0 | 127.0.0.1:9000 | voter | c3d485471729f2f04f6401530f7 |   UPGRADING   |
+--------+----------------+-------+-----------------------------+---------------+
| micro1 | 127.0.0.1:9001 | voter | 515c7eace0aba1f7a42761824ab | NEEDS UPGRADE |
+--------+----------------+-------+-----------------------------+---------------+
| micro2 | 127.0.0.1:9002 | voter | 4f461765de6659b0e683a5b11ed | NEEDS UPGRADE |
+--------+----------------+-------+-----------------------------+---------------+
| micro3 | 127.0.0.1:9003 | voter | d94955e2c26cb2e987489933bdz | NEEDS UPGRADE |
+--------+----------------+-------+-----------------------------+---------------+
```

The test now checks if all nodes report `ONLINE` after the first node updates. If they don't, it infers a schema update and attempts to update the remaining nodes. It finally checks that each node status reports `ONLINE` and this time fails the test if they don't.

If a node _does_ report all nodes as `ONLINE` before every node has updated, we can assume an update occurred that did not include a dqlite schema update. So we progress to check the ovsdb schema like before.

tests/test_helper/bats/ovsdb_schema_upgrade.bats
```diff
@@ -61,20 +61,47 @@ setup() {
...
+
+            wait_ovn_services "$container"
+            wait_microovn_online "$container" 30
+        fi
 
         # break out of the loop if every container has been upgraded
-        if [ $(( "$container_index" + 1 )) -eq  "$containers_size" ]; then
-            break
+        if [ $(( "$container_index" + 1 )) -eq  "$containers_size" ] || [ "$upgraded_size" -eq "$containers_size" ]; then
+          break
```
mkalcok, line 102: If I'm reading this correctly, we will break out of the "main" loop if the previous "inner" loop, that you added, already  upgraded all the hosts, correct? That makes sense as we can't perform the upgrade checks via API if the dqlite schema is changing. However there's this block of code at the end of this main loop
```
# check that the schema upgrade was not triggered yet.
        local current_sb=""
        local current_nb=""
        current_nb=$(get_current_ovsdb_schema_version "$probe_container" "nb")
        current_sb=$(get_current_ovsdb_schema_version "$probe_container" "sb")
        assert [ "$current_nb" == "$original_nb" ]
        assert [ "$current_sb" == "$original_sb" ]
```
that verifies that the OVN schema change was not triggered until every node is upgraded. These function do not rely on the microovn API and therefore they could (and should) be executed after every host upgrade even if the dqlite schema is changing.

nitpick: your additions in this file have inconsistent indentation (2 vs 4 spaces).

mkalcok (changes requested):
Thank you for this comprehensive change. It looks great. I have only one requested that I included inline. I'll take it for the spin on Monday and after that, I think we are good to go here :)

tests/test_helper/bats/ovsdb_schema_upgrade.bats
masnax: Good point, I've moved that check up into the first loop as well, and it will run after each system upgrades, except for the last one. On that one, we ensure that all nodes are ONLINE, and then check that the ovsdb schema update has been completed.

masnax:
@mkalcok docs have been updated now.

mkalcok (approved):
Thank you for getting this change in @gabrielmougard @masnax.  I know it's been a long journey, but I believe it was worth it. LGTM

gabrielmougard:
@mkalcok @masnax thanks a lot!

PR #145, opened 2024-06-28
Better error message for cluster add

crypticC0der:
Fix for microovn bug #2039578
The error message now tells the user in a much more user friendly manner that the name is taken when trying to add a node with a non-unique error name

### DIFF
--- a/microovn/cmd/microovn/cluster_add.go
+++ b/microovn/cmd/microovn/cluster_add.go
@@ -35,7 +35,9 @@ func (c *cmdClusterAdd) Run(cmd *cobra.Command, args []string) error {
 
 	token, err := m.NewJoinToken(args[0])
 	if err != nil {
-		return err
+		return fmt.Errorf("Unable to add server to MicroCluster, name %q is taken:\n%w",
+			args[0],
+			err)
 	}
 
 	fmt.Println(token)


microovn/cmd/microovn/cluster_add.go
```diff
@@ -35,7 +35,7 @@ func (c *cmdClusterAdd) Run(cmd *cobra.Command, args []string) error {
 
 	token, err := m.NewJoinToken(args[0])
 	if err != nil {
-		return err
+		return fmt.Errorf("Unable to add MicroCluster, name %q is taken",args[0])
```
fnordahl, line 38: Now the error is completely hidden, should we include it at the end of the string after a `: `?

Would the text improve if we add the word "server", i.e. "... add server to Micr ...". (Personally I would actually prefer "node" but I see the help text uses the word "server").

`gofmt -d .` also has opinions on this line. (We really should get the `make check-lint` and our CI to detect those for us.)

fnordahl (changes requested):
Thank you for picking this up, UX is really important to us!

I have a comment in-line and in addition to that I have some feedback on the structure of the commit message:
* If you look at the other commit messages in this part of the tree they tend to be prefixed with a short prefix detailing which part of the code is being updated.
* It's really great that you include a bug reference! I'd suggest putting it at the end of the message though. For consistency we can also use the format other projects use for Launchpad bugs, for example `Closes-Bug: ...`
* If you look at your commit message with `git log` in a 80x24 terminal the lines are too long.

crypticC0der:
modified the commit to address the issues noted

PR #146, opened 2024-06-28
Removed old version package

crypticC0der:
Fix for microovn bug [#2030489](https://bugs.launchpad.net/ubuntu/+source/ovn/+bug/2030489)

The static 0.1 version package has now been removed This makes the cobra command use the new ovn version and in the case of microovnd, the build system is modified to also include a build time variable of the git version which is used in the cobra command


mkalcok (approved):
LGTM, thank you.

mkalcok:
And of course, only after merging this change I noticed that the linked LP issue in the commit message is wrong. :facepalm: my bad, but I don't think it's worth amending it now.

PR #147, opened 2024-07-10
Update MicroCluster dependency

gabrielmougard:
This dependency bump brings https://github.com/canonical/microcluster/commit/3d587817830c05c0131a2deeedcd44ddc7ccf816 which will allow other MicroCluster based app to publicly expose their own API extensions through the status API that can be queried even before the MicroCluster database is initialized. This is particularly useful in the MicroCloud init phase, to check that an API extension feature is present before setting up special MicroOVN behavior (e.g, custom encapsulation IP for Geneve tunnel)


mkalcok (approved):
LGTM, thank you.

PR #149, opened 2024-07-11
Create devcontainer.json

xuhao61:


mkalcok:
What is the purpose of this change? Generally we'd expect commit message to contain details about the change (see [CONTRIBUTING.md](https://github.com/canonical/microovn/blob/main/CONTRIBUTING.md)). As far as I can tell `devcontainer.json` is a feature specific to some IDEs or development workflows, and as such I don't think it belongs to the repository.

PR #151, opened 2024-07-16
Fix for flapping ovsdb schema upgrade test

crypticC0der:
Testsuite 'ovsdb_schema_upgrade.bats' was unstable and occasionally failed, this commit fixes that by modifying the wait_microovn_online to call a function that waits until each nodes status is valid given a regex string.

This function is used to check if node is upgrading, needing an upgrade or online instead of just online. This allows the test to pass reliably on my machine.

Reported-at: https://bugs.launchpad.net/microovn/+bug/2072658


fnordahl (approved):
LGTM, thanks!

PR #152, opened 2024-07-24
Added the ability to disable/enable services

crypticC0der:
Adds commands and testing for the disabling and enabling of services, this is done through microovn enable <service> and microovn disable <service>.

This is deemed useful as the inability to control what nodes important services, such as central, are running on can mean a user may have their central database on lower performance nodes without the ability to move them.

### DIFF
--- a/microovn/api/endpoints.go
+++ b/microovn/api/endpoints.go
@@ -19,6 +19,7 @@ var Server = rest.Server{
 			PathPrefix: types.APIVersion,
 			Endpoints: []rest.Endpoint{
 				services.ListCmd,
+				services.ServiceControlCmd,
 				certificates.IssueCertificatesEndpoint,
 				certificates.IssueCertificatesAllEndpoint,
 				certificates.RegenerateCaEndpoint,
--- a/microovn/api/endpoints.go
+++ b/microovn/api/endpoints.go
@@ -6,6 +6,7 @@ import (
 	"github.com/canonical/microovn/microovn/api/ovsdb"
 
 	"github.com/canonical/microovn/microovn/api/certificates"
+	"github.com/canonical/microovn/microovn/api/services"
 	"github.com/canonical/microovn/microovn/api/types"
 )
 
@@ -16,7 +17,8 @@ var Server = rest.Server{
 	Resources: []rest.Resources{
 		{
 			PathPrefix: types.APIVersion,
-			Endpoints: []rest.Endpoint{servicesCmd,
+			Endpoints: []rest.Endpoint{
+				services.ListCmd,
 				certificates.IssueCertificatesEndpoint,
 				certificates.IssueCertificatesAllEndpoint,
 				certificates.RegenerateCaEndpoint,
--- a/microovn/api/services/control.go
+++ b/microovn/api/services/control.go
@@ -1,0 +1,51 @@
+package services
+
+import (
+	"errors"
+	"net/http"
+	"net/url"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/state"
+	"github.com/gorilla/mux"
+
+	"github.com/canonical/microovn/microovn/ovn"
+)
+
+// /1.0/services/service endpoint.
+var ServiceControlCmd = rest.Endpoint{
+	Path:   "service/{service}",
+	Delete: rest.EndpointAction{Handler: disableService, AllowUntrusted: false, ProxyTarget: true},
+	Put:    rest.EndpointAction{Handler: enableService, AllowUntrusted: false, ProxyTarget: true},
+}
+
+func enableService(s *state.State, r *http.Request) response.Response {
+	requestedService, err := url.PathUnescape(mux.Vars(r)["service"])
+	if err != nil {
+		return response.InternalError(err)
+	}
+	if !ovn.CheckValidService(requestedService) {
+		return response.InternalError(errors.New("Service does not exist"))
+	}
+	err = ovn.EnableService(s, requestedService)
+	if err != nil {
+		return response.InternalError(err)
+	}
+	return response.SyncResponse(true, requestedService+" enabled")
+}
+
+func disableService(s *state.State, r *http.Request) response.Response {
+	requestedService, err := url.PathUnescape(mux.Vars(r)["service"])
+	if err != nil {
+		return response.InternalError(err)
+	}
+	if !ovn.CheckValidService(requestedService) {
+		return response.InternalError(errors.New("Service does not exist"))
+	}
+	err = ovn.DisableService(s, requestedService)
+	if err != nil {
+		return response.InternalError(err)
+	}
+	return response.SyncResponse(true, requestedService+" disabled")
+}
--- a/microovn/ovn/services.go
+++ b/microovn/ovn/services.go
@@ -1,0 +1,90 @@
+package ovn
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"slices"
+
+	"github.com/canonical/microcluster/state"
+	"github.com/canonical/microovn/microovn/database"
+	"github.com/canonical/microovn/microovn/node"
+)
+
+type SrvName string
+
+const (
+	SrvChassis SrvName = "chassis"
+	SrvCentral SrvName = "central"
+	SrvSwitch  SrvName = "switch"
+)
+
+var ServiceNames = []SrvName{SrvChassis, SrvCentral, SrvSwitch}
+
+func CheckValidService(service string) bool {
+	return slices.Contains(ServiceNames, SrvName(service))
+}
+
+func DisableService(s *state.State, service string) error {
+	exists, err := node.HasServiceActive(s, service)
+
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return errors.New("This service is not enabled")
+	}
+
+	if SrvName(service) == SrvCentral {
+		err = snapStop("ovn-ovsdb-server-nb", true)
+		if err != nil {
+			return err
+		}
+		err = snapStop("ovn-ovsdb-server-sb", true)
+		if err != nil {
+			return err
+		}
+		err = snapStop("ovn-northd", true)
+	} else {
+		err = snapStop(service, true)
+	}
+
+	if err != nil {
+		return fmt.Errorf("Snapctl error, likely due to service not existing:\n %w", err)
+	}
+
+	err = s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
+		err := database.DeleteService(ctx, tx, s.Name(), service)
+		return err
+	})
+
+	return err
+
+}
+
+func EnableService(s *state.State, service string) error {
+	exists, err := node.HasServiceActive(s, service)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return errors.New("This Service is already enabled")
+	}
+
+	if !CheckValidService(service) {
+		return errors.New("Service does not exist")
+	}
+	err = snapStart(service, true)
+	if err != nil {
+		return fmt.Errorf("Snapctl error, likely due to service not existing:\n%w", err)
+	}
+
+	err = s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
+		_, err := database.CreateService(ctx, tx, database.Service{Member: s.Name(), Service: service})
+		return err
+	})
+
+	return err
+
+}
--- a/tests/test_helper/bats/services.bats
+++ b/tests/test_helper/bats/services.bats
@@ -1,0 +1,61 @@
+
+load "${ABS_TOP_TEST_DIRNAME}test_helper/setup_teardown/$(basename "${BATS_TEST_FILENAME//.bats/.bash}")"
+
+setup() {
+    load ${ABS_TOP_TEST_DIRNAME}test_helper/common.bash
+    load ${ABS_TOP_TEST_DIRNAME}test_helper/lxd.bash
+    load ${ABS_TOP_TEST_DIRNAME}test_helper/microovn.bash
+    load ${ABS_TOP_TEST_DIRNAME}../.bats/bats-support/load.bash
+    load ${ABS_TOP_TEST_DIRNAME}../.bats/bats-assert/load.bash
+
+    # Ensure TEST_CONTAINERS is populated, otherwise the tests below will
+    # provide false positive results.
+    assert [ -n "$TEST_CONTAINERS" ]
+}
+
+services_register_test_functions() {
+    bats_test_function \
+        --description "Testing of service functionality" \
+        -- service_tests
+}
+
+service_tests() {
+    for container in $TEST_CONTAINERS; do
+        # enable enabled service
+        run lxc_exec "$container" "microovn enable switch"
+        assert_output "Error: Failed to enable service 'switch': 'This Service is already enabled'"
+
+        # enable non existing service
+        run lxc_exec "$container" "microovn enable switchh"
+        assert_output "Error: Failed to enable service 'switchh': 'Service does not exist'"
+
+        # disable non existing service
+        run lxc_exec "$container" "microovn disable switchh"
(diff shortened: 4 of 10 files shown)


microovn/ovn/services.go
```diff
@@ -0,0 +1,69 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/canonical/microcluster/state"
+	"github.com/canonical/microovn/microovn/database"
+)
+
+func DisableService(s *state.State, service string) error {
+	err := s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
+		exists, err := database.ServiceExists(ctx, tx, s.Name(), service)
```
fnordahl, line 15: The `service` string is supplied by the end user of the program, we need to do some sort of validation of it before passing it on into the innards of the program.

Who knows what the user could craft to have the database or `snapctl` do unexpected things with elevated privileges.

microovn/ovn/services.go
```diff
@@ -0,0 +1,69 @@
...
+		if exists != true {
+			return errors.New("No such service")
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	err = snapStop(service, true)
```
fnordahl, line 28: Some of our services, such as `chassis` and `switch` are simple and map directly to a single systemd service.

However we do also have more complex services, such as `central` which in reality map into multiple systemd services:
https://github.com/canonical/microovn/blob/a99570ddbcffe6acbd76698a424c78d3cca49773/microovn/ovn/join.go#L113-L126

This is a useful abstraction and we want to keep the components individually controllable at the lower level.

However, for the purpose of the `microovn enable` and `microovn disable` commands we need some sort of abstraction so that the user can do `microovn disable central` and have the system disable all the underlying systemd services that compose that higher level service.

microovn/api/endpoints.go
```diff
@@ -16,7 +17,9 @@ var Server = rest.Server{
 	Resources: []rest.Resources{
 		{
 			PathPrefix: types.APIVersion,
-			Endpoints: []rest.Endpoint{servicesCmd,
```
fnordahl, line 19: Making the existing `servicesCmd` function public and renaming the module is a logical unit of change in itself. Please put that into a separate commit so that it is easier to discern between reorganization/refactoring of existing code and new code during review.

fnordahl (changes requested):
Hello, MJ, and thank you for your work on this!

The approach overall looks good to me and there are some comments in-line requesting some changes.

In addition we would like to see some documentation for this new feature.

I'll defer to @mkalcok for any additional comments on implementation.

microovn/ovn/services.go
```diff
@@ -0,0 +1,69 @@
...
+		err := database.DeleteService(ctx, tx, s.Name(), service)
+		return err
+	})
+
+	return err
+
+}
+
+func EnableService(s *state.State, service string) error {
+	err := s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
```
mkalcok, line 43: There's a function that might give you the same functionality as this transaction
https://github.com/canonical/microovn/blob/a99570ddbcffe6acbd76698a424c78d3cca49773/microovn/node/services.go#L41-L43

tests/test_helper/bats/services.bats
```diff
@@ -0,0 +1,80 @@
...
+    for container in $TEST_CONTAINERS; do
+        run lxc_exec "$container" "microovn disable switch"
+        assert_output "Service switch disabled"
+    done
+}
+
+disable_non_enabled_service() {
+    for container in $TEST_CONTAINERS; do
+        run lxc_exec "$container" "microovn disable switch"
+        assert_output "Error: command failed: No such service"
```
mkalcok, line 54: I would suggest improving output of the command in this case (also in `enable_already_enabled_service`). Instead of `No such service`, which isn't really descriptive, it could say `Service <service_name> already disabled`. I'd even avoid all the `Error` and `command failed` prefixes.

tests/test_helper/bats/services.bats
```diff
@@ -0,0 +1,80 @@
...
+    load ${ABS_TOP_TEST_DIRNAME}test_helper/microovn.bash
+    load ${ABS_TOP_TEST_DIRNAME}../.bats/bats-support/load.bash
+    load ${ABS_TOP_TEST_DIRNAME}../.bats/bats-assert/load.bash
+
+    # Ensure TEST_CONTAINERS is populated, otherwise the tests below will
+    # provide false positive results.
+    assert [ -n "$TEST_CONTAINERS" ]
+}
+
+services_register_test_functions() {
```
mkalcok, line 16: These tests seem to be heavily dependent on order of execution which is not the best practice. If forced order of operations is really necessary, it should be a single test. Otherwise each test should restore state of the system on teardown.

As it may be a bit hard to restore system state on failure, I'd perhaps suggest to go with single long test here that bails in case of any failure.

tests/test_helper/bats/services.bats
```diff
@@ -0,0 +1,80 @@
...
+    bats_test_function \
+        --description "Disable non enabled service" \
+        -- disable_non_enabled_service
+    bats_test_function \
+        --description "Enable disabled service" \
+        -- enable_disabled_service
+}
+
+enable_disabled_service() {
+    for container in $TEST_CONTAINERS; do
```
mkalcok, line 38: suggestion: I'd say that it's enough to run the operation on single node. Is there a value in executing it on every node?

tests/test_helper/setup_teardown/basic_cluster.bash
```diff
@@ -4,7 +4,7 @@ setup_file() {
     load test_helper/microovn.bash
 
 
-    TEST_CONTAINERS=$(container_names "$BATS_TEST_FILENAME" 3)
+    TEST_CONTAINERS=$(container_names "$BATS_TEST_FILENAME" 1)
```
mkalcok, line 7: Why was this changed? `basic_cluster.bats` should operate on 3 node cluster.

mkalcok (changes requested):
The content overall looks good to me, thank you MJ. I left couple of suggestions inline that should be considered.

crypticC0der:
I have addressed the suggestions by @fnordahl and will implement the ones submitted by martin

microovn/api/services/control.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// /1.0/services/service endpoint.
+var ServiceControlCmd = rest.Endpoint{
+	Path:   "service/{service}",
+	Delete: rest.EndpointAction{Handler: disableService, AllowUntrusted: false, ProxyTarget: true},
+	Put:    rest.EndpointAction{Handler: enableService, AllowUntrusted: false, ProxyTarget: true},
+}
+
+func enableService(s *state.State, r *http.Request) response.Response {
```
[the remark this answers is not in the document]
crypticC0der: I don't see why we would need to validate the input twice, that sounds unnecessary

microovn/ovn/services.go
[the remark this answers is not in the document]
crypticC0der: it uses the snap start and snap stop functions

microovn/api/services/control.go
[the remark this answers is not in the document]
mkalcok: If we manage to have single source of truth for these checks I don't think it's necessarily bad to validate input in multiple places. I think it makes sense to:
* Validate in request handler to stop processing invalid request as soon as possible
* Validate in function that actually implements the enable/disable functionality because there might be multiple code paths leading to that function call.

Though if you decide to go with one validation, I agree that the second one is more important. Just make sure that the invalid can not be misused for something else in the request handler.

microovn/cmd/microovn/disable.go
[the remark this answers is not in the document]
crypticC0der: yes I agree

microovn/cmd/microovn/disable.go
```diff
@@ -0,0 +1,83 @@
...
+
+type cmdDisable struct {
+	common  *CmdControl
+	cluster *cmdCluster
+}
+
+func (c *cmdDisable) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "disable <SERVICE>",
+		Short: "disables a service",
```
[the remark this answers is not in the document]
crypticC0der: to be discussed tomorrow

microovn/ovn/services.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	SrvChassis SrvName = "chassis"
+	SrvCentral SrvName = "central"
+	SrvSwitch  SrvName = "switch"
+)
+var ServiceNames = []SrvName{SrvChassis,SrvCentral,SrvSwitch}
+
+func DisableService(s *state.State, service string) error {
+	err := s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
```
fnordahl, line 24: We need to do validation of the `service` string prior to this call.

microovn/ovn/services.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+
+	if SrvName(service) == SrvCentral{
+		err = snapStop("ovn-ovsdb-server-nb",true)
+		if err != nil {return err }
+		err = snapStop("ovn-ovsdb-server-sb",true)
+		if err != nil {return err }
+		err = snapStop("ovn-northd",true)
+	}
```
fnordahl, line 44: Nice!

While there exists a `central` systemd service in addition to the above ones, it is kept as a transitional service to avoid breaking anyone's workflow (see commit 595567eb1d33016a30ef76cb98433250ae4358b0).

For our purposes we should avoid using it, so let's add a `else` branch for the case of `service == SrvCentral`.

microovn/ovn/services.go
```diff
@@ -0,0 +1,84 @@
...
+		err := database.DeleteService(ctx, tx, s.Name(), service)
+		return err
+	})
+
+	return err
+
+}
+
+func EnableService(s *state.State, service string) error {
+	exists, err := node.HasServiceActive(s, service)
```
fnordahl, line 61: We need to do validation of the service string prior to this call.

microovn/cmd/microovn/disable.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+
+	cli, err := m.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	targetService := args[0]
```
fnordahl, line 39: This will crash if no arguments are passed, i.e. if the user enters `microovn disable` without actually passing a service name.

microovn/cmd/microovn/disable.go
```diff
@@ -0,0 +1,83 @@
...
```
fnordahl, line 75: This will crash if no arguments are passed, i.e. if the user enters `microovn enable` without actually passing a service name.

fnordahl (dismissed):
In addition to the in-line comments `gofmt -d ./..` has opinions on formatting of some of the code.

I took this for a spin and while I am able to enable the central services on a fourth node using the `microovn enable central` command, there is currently no handling of actually making those services a part of the OVSDB clusters.

I realize that we have not yet specced that out, so I wonder whether to address that in this PR or if we should merge as-is and tackle that as the next challenge in the coming pulse. Thoughts?

tests/test_helper/bats/services.bats
```diff
@@ -0,0 +1,50 @@
...
+        run lxc_exec "$container" "microovn enable switchh"
+        assert_output "Error: Failed to enable service 'switchh': 'Service does not exist'"
+
+        # disable non existing service
+        run lxc_exec "$container" "microovn disable switchh"
+        assert_output "Error: Failed to disable service 'switchh': 'This service is not enabled'"
+
+        # disable service
+        run lxc_exec "$container" "microovn disable switch"
+        assert_output "Service switch disabled"
```
mkalcok, line 38: I forgot to mention this in my last review, but I think that the test should verify more than just command output. E.g:
* Check `microovn status` output to make sure that it reports expected services on the node
* Check `snap services microovn` to make sure that expected services were actually started

mkalcok:
> I realize that we have not yet specced that out, so I wonder whether to address that in this PR or if we should merge as-is and tackle that as the next challenge in the coming pulse. Thoughts?

We could limit this feature to `switch` and `chassis` services for now, and add support for `central` in next pulse.

microovn/ovn/services.go
crypticC0der: why, the service name gets validated within this call by checking it against the database?

fnordahl: It is a precaution in the event of vulnerable code in our library dependencies.

Our backing database is SQL-based, and history is riddled with security incidents rooted in SQL injection attacks.

While our code has solid mitigations for that by making use of [prepared statements](https://go.dev/doc/database/prepared-statements) with place holders for variables:
https://github.com/canonical/microovn/blob/a99570ddbcffe6acbd76698a424c78d3cca49773/microovn/database/service.mapper.go#L51-L55

which should avoid such issues, I'd rather be safe than sorry by validating the data before exposing it to our data store.

crypticC0der: I did look and check that the input is sanitized for the database queries, I shall validate the input in the API, though I stand by my opinion on it being somewhat unneccicary

fnordahl:
> we have support for central, it can be enabled, which enables its child services due to a snap hook, and disabled which disables its child services directly

Indeed, but when you enable the central services on a fresh node for the first time. Does the clustered databases become a member of the cluster?

crypticC0der:
> > we have support for central, it can be enabled, which enables its child services due to a snap hook, and disabled which disables its child services directly
> 
> Indeed, but when you enable the central services on a fresh node for the first time. Does the clustered databases become a member of the cluster?

okay thats a good point, I am unsure,

mkalcok:
> it can be enabled, which enables its child services due to a snap hook

I'd really discourage the usage of  the `central` snap service. I think we should remove it for the `24.03` release as it's just a source of confusion.

crypticC0der:
all that is left on this is
documentation, a list command for the services, a decision on what to do regarding central seemingly
all other issues have been fixed, and more tests have been added

crypticC0der:
I have sorted out formatting now also

fnordahl (commented):
From the perspective of enabling/disabling services functionality, this looks good to me. If we can have a commit documenting the new commands we should be ready.

For the `central` service there is additional work required to handle OVSDB server join and departure events which we can do as a follow-up if you'd like.

tests/test_helper/bats/services.bats
```diff
@@ -0,0 +1,61 @@
...
+
+        run lxc_exec "$container" "snap services microovn | grep switch | grep enabled"
+        assert_output ""
+
+        # enable disabled service
+        run lxc_exec "$container" "microovn enable switch"
+        assert_output "Service switch enabled"
+
+        assert [ -n "$(run lxc_exec "$container" "microovn status | grep switch")"]
+        assert [ -n "$(run lxc_exec "$container" "snap services microovn | grep switch | grep enabled")"]
```
mkalcok, line 55: Couple of notes here:

* We could use the same sets of checks after service is disabled (just with the logic inverted)
* Wouldn't `microovn status | grep switch` produce false positives if the `switch` service is enabled on a different node?
* nit: `run` is a [function from BATS](https://bats-core.readthedocs.io/en/stable/writing-tests.html#run-test-other-commands) that stores output of the command in `$output` and return code in `$status`. More idiomatic way to approach this would be to use `run` in one statement and then do assert on the `$output`, or since `grep` will fail if it does not find a match, you could `run` and then [`assert_success`](https://github.com/bats-core/bats-assert?tab=readme-ov-file#assert_success)

mkalcok (dismissed):
I have last few suggestions that I left in-line. Also, we don't have this stated outright in our `CONTRIBUTING.md`, but I think we should keep documenting code in [Go Doc style](https://go.dev/doc/comment), at least for functions and types. (wdyt @fnordahl ?)

tests/test_helper/bats/services.bats
crypticC0der: there is only one node in this test

mkalcok (approved):
LGTM, thank you.

mkalcok (approved):

PR #153, opened 2024-07-24
tests: Fix for flaky ovsdb_schema_upgrade test

mkalcok:
please review each commit individually.

### DIFF
--- a/tests/test_helper/bats/ovsdb_schema_upgrade.bats
+++ b/tests/test_helper/bats/ovsdb_schema_upgrade.bats
@@ -63,24 +63,31 @@ setup() {
     local containers_size="${#containers_to_upgrade[@]}"
     local upgraded_size=0
 
+    # Loop below detects and performs internal dqlite schema upgrade if necessary.
+    # When upgrading dqlite, all nodes have to be upgraded before cluster becomes
+    # usable again.
     for ((container_index = 0; container_index < "$containers_size"; container_index++)) do
         local container="${containers_to_upgrade[$container_index]}"
         # Upgrade MicroOVN cluster, one cluster member at a time.
         install_microovn "$MICROOVN_SNAP_PATH" "$container"
         upgraded_size=$(( "$upgraded_size" + 1 ))
 
-        # Check if the node is fully online. If it's not, then we have a dqlite schema upgrade and need to upgrade all nodes first.
-        # If all nodes are reporting ONLINE, then we can move on to check the ovsdb schema before upgrading all nodes.
-        wait_ovn_services "$container"
-        run wait_microovn_online "$container" 30
-        if [ "$status" = 0 ] ; then
-            echo "# ($container) reports all systems are reachable. No dqlite schema update detected"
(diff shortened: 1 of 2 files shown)


fnordahl:
I'm happy to merge this and I'd like @crypticC0der's review of it prior to doing so.

mkalcok:
For the record, I manually triggered the `ovsdb_schema_upgrade` test couple of times in the CI and I see it consistently passing. Previously this test would fail about 50% of the times

tests/test_helper/microovn.bash
```diff
@@ -381,19 +381,19 @@ function microovn_wait_for_service_starttime() {
 # If cluster members do not reach "online" state before the MAX_RETRY is reached, this
 # function returns 1 as a return code.
 
-function wait_microovn_state(){
```
crypticC0der, line 384: unsure as to the reasoning for removing this generic function, realistically it could be useful, though I am biased on this.

tests/test_helper/bats/ovsdb_schema_upgrade.bats
```diff
@@ -63,24 +63,31 @@ setup() {
...
+            if [ "$status" = 0 ] ; then
+                echo "# ($container) reports all systems are reachable. No dqlite schema update detected" >&3
+                break
+            fi
         fi
 
         # If we upgraded all systems because of a dqlite schema update, ensure all nodes are online.
         if [ "$upgraded_size" -eq "$containers_size" ] ; then
-            grep -q 0 <<< "$status"
+            run wait_microovn_online "$container" 45
```
crypticC0der, line 89: we are sure that 45s timeout is proving to be enough time, correct?

crypticC0der (approved):
Really good Martin, I see how this works more reliably

tests/test_helper/microovn.bash
mkalcok: It wasn't anything personal :laughing: but there were some issues with `wait_microovn_{state,upgrading,online}`
* `wait_microovn_upgrading` was matching on `ONLINE` state as well, which means that it would return `true` when executed against fully online cluster
* `wait_microovn_online` and `wait_microovn_upgrading` were missing documentation
* `wait_microovn_state` documentation was outdated

And since the `wait_microovn_upgrading` became unused, I took the easy way out and removed it instead of fixing above issues.

tests/test_helper/bats/ovsdb_schema_upgrade.bats
mkalcok: I gotta admit, I'm still not fully clear on why was this occasionally failing. When trying manually, I see cluster transition form `UPGRADING/NEEDS UPGRADE` to `ONLINE` almost instantly, every time. Nevertheless, i believe this new approach is sufficient for us as it explicitly tests that after upgrading every node, they become online within 45 seconds, and it looks to be stable in CI.

fnordahl (approved):
LGTM, thanks both!

PR #154, opened 2024-07-29
tests: Increase upgrade timeout

mkalcok:
Due to the `microcluster` version used in `22.03` release, it can sometimes take two (30 seconds) heartbeats to fully converge cluster after upgrade. This was fixed in later versions [0], but we are increasing timeout to accommodate tests of older versions.

[0] https://github.com/canonical/microcluster/pull/113/commits/d78347458c4a3efb38cf86c5cfb8f26355b8093f


crypticC0der (approved):
many such cases

PR #155, opened 2024-07-30
Isolation of snap and service code within ovn

crypticC0der:
This removes certain parts of the ovn package codebase for controlling services and interacting with snap, and moves them to their own package. This was needed as the ovn package was seemingly getting bloated and this strips it down a bit


fnordahl (commented):
Hey, MJ, thanks for the patch!

This change appears to move two modules, `snap` and `services`. I think that calls for two separate commits as well to contain the two logical units of change?

The reason why this is important is that you never know what the future holds, perhaps we want to cherry pick one of the changes to a stable branch, or revert one of the changes, or maybe just have a clear explanation for why something was changed on `git blame`.

mkalcok:
> I think that calls for two separate commits as well to contain the two logical units of change?

@fnordahl does it apply even when the changes are depending on each other? Because without moving snap-related functions to the `snap` package, contents of the `services` would have to remain in the `ovn` package.

Though I didn't do a full review yet, I have few notes:

Is it necessary to rename `node` package to `services`? I thought that the original package name was rather semantically fitting. It had functions like `HasService` which naturally translated to calls like `node.HasService()`, answering question "Does this node has this service". Even with your additions of `DisableService` and `EnableService`, calls like `node.EnableService()` (enable service on this node) would make a lot of sense.
Current implementation feels bit stuttery `services.HasService()`.

Now that the `snap` package is isolated, functions in this package could lose the `snap` prefix, otherwise it again leads to a lot of stuttering with calls like `snap.snapStart()`.

fnordahl:
> @fnordahl does it apply even when the changes are depending on each other? Because without moving snap-related functions to the `snap` package, contents of the `services` would have to remain in the `ovn` package.

While I have not dived into the the details, I see one group of files changed referring to only one of the modules and another referring to the other module, so presumably those changes can be made in separate commits?

crypticC0der:
wow these commits are now in two!!!!!

crypticC0der:
> Now that the snap package is isolated, functions in this package could lose the snap prefix, otherwise it again leads to a lot of stuttering with calls like snap.snapStart().
I entirely agree on this

> Is it necessary to rename node package to services?
I dont think so, but I think that the service related functionality should be in the same place, so either in node or service

crypticC0der:
I have made a commit changing the snap commands to 
snap.Snap* to just snap.*

mkalcok:
> I dont think so, but I think that the service related functionality should be in the same place, so either in node or service

I agree with pulling of the `{Enable,Disable}Service` functions out of the `ovn` package, but if we are going to rename existing packages, there needs to be a good reason for it.

crypticC0der:
@mkalcok should be all good now

microovn/snap/snap.go
```diff
@@ -26,7 +26,7 @@ func snapStart(service string, enable bool) error {
 
 // snapStop stops specified snap service. Service can be optionally also disabled, ensuring
 // that it won't be automatically started on system reboot.
-func snapStop(service string, disable bool) error {
+func Stop(service string, disable bool) error {
```
mkalcok, line 29: GoDoc does not match the renamed function

microovn/node/node.go
mkalcok, line 1: Name of the file doesn't matter that much, since the package name remains `node`, though I'm still wondering if the file name change was necessary. Previous logic behind naming was that this package holds functions related to specific MicroOVN node and the file `services` would sub-group functions for services on the specific node.

mkalcok (changes requested):
Overall looks good, thank you. I have but one last request, since we will be enabling linter that ensures GoDocs, could you please add them for functions/structures that you recently added (and moved in this change)?
* SrvName
* CheckValidService
* DisableService
* EnableService

mkalcok (approved):
Thank you, LGTM.

PR #157, opened 2024-08-01
Quality of life changes for service migration

crypticC0der:
Implements various warnings for the Service Migration within microovn,
This includes:
- a warning on an even number of microovn central nodes
- a warning on less than 3 microovn central nodes
- an error on disabling the final central service

There are also a new set of tests for ensuring this functionality works

### DIFF
--- a/microovn/api/services/control.go
+++ b/microovn/api/services/control.go
@@ -6,10 +6,12 @@ import (
 	"net/url"
 
 	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microcluster/v2/rest"
 	"github.com/canonical/microcluster/v2/state"
 	"github.com/gorilla/mux"
 
+	"github.com/canonical/microovn/microovn/api/types"
 	"github.com/canonical/microovn/microovn/node"
 )
 
@@ -20,10 +22,16 @@ var ServiceControlCmd = rest.Endpoint{
 	Put:    rest.EndpointAction{Handler: enableService, AllowUntrusted: false, ProxyTarget: true},
 }
 
+// enableService - function to handle to service control put request,
+// which aims to enable a service.
+//
+// This will return a response which contains a WarningSet for the
+// current desired state and a response string on the operation
 func enableService(s state.State, r *http.Request) response.Response {
 	requestedService, err := url.PathUnescape(mux.Vars(r)["service"])
 	if err != nil {
-		return response.InternalError(err)
+		logger.Errorf("Failed to get service: %s", err)
+		return response.ErrorResponse(500, "Internal server error")
 	}
 	if !node.CheckValidService(requestedService) {
 		return response.InternalError(errors.New("Service does not exist"))
@@ -32,13 +40,28 @@ func enableService(s state.State, r *http.Request) response.Response {
 	if err != nil {
 		return response.InternalError(err)
 	}
-	return response.SyncResponse(true, requestedService+" enabled")
+
+	scr := types.ServiceControlResponse{}
+	scr.Warnings, err = node.ServiceWarnings(r.Context(), s)
+	if err != nil {
+		logger.Errorf("Failed to generate warnings for service: %s: %s", requestedService, err)
+		return response.ErrorResponse(500, "Internal server error")
+	}
+	scr.Message = requestedService + " enabled"
+
+	return response.SyncResponse(true, scr)
 }
 
+// disableService - function to handle to service control delete request,
+// which aims to disable a service.
+//
+// This will return a response which contains a WarningSet for the
+// current desired state and a response string on the operation
 func disableService(s state.State, r *http.Request) response.Response {
 	requestedService, err := url.PathUnescape(mux.Vars(r)["service"])
 	if err != nil {
-		return response.InternalError(err)
+		logger.Errorf("Failed to get service: %s", err)
+		return response.ErrorResponse(500, "Internal server error")
 	}
 	if !node.CheckValidService(requestedService) {
 		return response.InternalError(errors.New("Service does not exist"))
@@ -47,5 +70,14 @@ func disableService(s state.State, r *http.Request) response.Response {
 	if err != nil {
 		return response.InternalError(err)
 	}
-	return response.SyncResponse(true, requestedService+" disabled")
+
+	scr := types.ServiceControlResponse{}
+	scr.Warnings, err = node.ServiceWarnings(r.Context(), s)
+	if err != nil {
+		logger.Errorf("Failed to generate warnings for service: %s: %s", requestedService, err)
+		return response.ErrorResponse(500, "Internal server error")
+	}
+	scr.Message = requestedService + " disabled"
+
+	return response.SyncResponse(true, scr)
 }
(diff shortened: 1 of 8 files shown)


crypticC0der:
as a note, this does require the package migration from pr #155

microovn/cmd/microovn/service_control.go
```diff
@@ -47,6 +47,8 @@ func (c *cmdDisable) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	fmt.Printf("Service %s disabled\n", targetService)
+	warnings, err := client.GetServiceWarningsString(context.Background(), cli)
+	fmt.Printf(warnings)
```
mkalcok, line 51: Should these warnings go to `STDERR`?

microovn/client/client.go
```diff
@@ -148,11 +148,43 @@ func DisableService(ctx context.Context, c *client.Client, serviceName string) e
...
+	if err != nil {
+		return warnString.String(), err
+	}
+
+	if warnings.EvenCentral {
+		warnString.WriteString("Warning: There are an even number of central service nodes, this cannot tolerate any more failures than if you had one less central service node and just leads to decreased write performance\n")
+	}
+
+	if warnings.FewCentral {
+		warnString.WriteString("Warning: There are less than 3 central service nodes, therefore your system cannot tolerate a single node failure putting your system at risk, you should add more central service nodes\n")
```
mkalcok, line 187: These warnings are rather verbose. Could we make them shorter, focusing on the core issue? Examples:
```
[central] Warning: Cluster with even number of members has same fault tolerance, but higher quorum requirements, than cluster with one less member.
```
```
[central] Warning: Cluster with less than 3 nodes can't tolerate any node failures.
```

mkalcok (changes requested):
Thanks MJ, this feature will, overall, greatly improve user experience and prevent unintended misconfiguration. I left couple of notes in-line, and from a high level, I'd like to discuss the API design.

Is it necessary for this to be implemented as an standalone endpoint? I think that this feature could be integrated directly into the [`enable`](https://github.com/canonical/microovn/blob/994eb5e4490274e1a4b5c03725d8ba1905f92c90/microovn/api/services/control.go#L23) and [`disable`](https://github.com/canonical/microovn/blob/994eb5e4490274e1a4b5c03725d8ba1905f92c90/microovn/api/services/control.go#L38) endpoints by returning richer response. Currently these endpoints return plain text, but they could return, for example, something like:

```go
type ServiceControlResponse struct {
	Message string `json:"message" yaml:"message"`
	Warnings WarningSet `json:"warnings" yaml:"warnings"`
}
```

This approach would not increase number of existing endpoints and also decreased number of api calls needed when performing enable/disable operations from client.

Additional thing worth considering is whether we want to print these warning indiscriminately whenever user performs `enable`/`disable` action or, since these are `central`-related warnings, they would be printed only when user enables/disables central service.

microovn/cmd/microovn/service_control.go
crypticC0der: i dont think so, they make sense in stdout

microovn/api/services/control.go
```diff
@@ -32,7 +33,15 @@ func enableService(s *state.State, r *http.Request) response.Response {
 	if err != nil {
 		return response.InternalError(err)
 	}
-	return response.SyncResponse(true, requestedService+" enabled")
+
+	scr := types.ServiceControlResponse{}
+	scr.Warnings, err = node.ServiceWarnings(s)
+	if err != nil {
+		return response.InternalError(err)
```
mkalcok, line 40: A good defensive practice in APIs is to not expose cause of internal server error to the caller. The error should be logged instead (preferably with some added context) and API caller should receive only generic error message. Example:
https://github.com/canonical/microovn/blob/803dc03ea847d214941330c74cc0ae0027550dc0/microovn/api/certificates/issue.go#L31-L34

microovn/node/node.go
```diff
@@ -40,6 +40,13 @@ func DisableService(s *state.State, service string) error {
...
+		centrals, err := FindService(s,service)
+		if err!=nil {
+			return err
+		}
+		if len(centrals)==1 {
+			return errors.New("You cannot delete the final enabled central service")
+		}
 		if err != nil {
 			return err
 		}
```
mkalcok, line 52: Isn't this err handling superfluous? There doesn't seem to be anything else that could produce error after the `FindService` call and that's already handled.

microovn/node/node.go
```diff
@@ -40,6 +40,13 @@ func DisableService(s *state.State, service string) error {
 
 	if SrvName(service) == SrvCentral {
 		err = snap.Stop("ovn-ovsdb-server-nb", true)
+		centrals, err := FindService(s,service)
+		if err!=nil {
```
mkalcok, line 44: nit: spaces between operands and operator

microovn/node/node.go
```diff
@@ -40,6 +40,13 @@ func DisableService(s *state.State, service string) error {
...
+			return err
+		}
+		if len(centrals)==1 {
```
mkalcok, line 47: nit: spaces between operands and operator

microovn/node/node.go
```diff
@@ -173,3 +180,18 @@ func FindService(s *state.State, service string) ([]cluster.InternalClusterMembe
 
 	return membersWithService, nil
 }
+
+func ServiceWarnings(s *state.State) (types.WarningSet, error) {
```
mkalcok, line 184: could you please add GoDoc for this function?

microovn/node/node.go
```diff
@@ -173,3 +180,18 @@ func FindService(s *state.State, service string) ([]cluster.InternalClusterMembe
...
+	output := types.WarningSet{}
+	centrals, err := FindService(s,string(SrvCentral))
+	if err!=nil {
```
mkalcok, line 187: nit: spaces between operands and operator

mkalcok (changes requested):
Just a couple of nits I mentioned in-line, otherwise LGTM :)

microovn/api/types/services.go
```diff
@@ -1,13 +1,47 @@
...
+	// Message - any output needed from the service control functions.
+	Message string `json:"message" yaml:"message"`
+	// Warnings - the set of warnings with the desired state of services.
+	Warnings WarningSet `json:"warnings" yaml:"warnings"`
+}
+
+// PrettyPrint - Formats and prints contents of WarningSet object.
+func (w WarningSet) PrettyPrint() {
+	if w.EvenCentral {
+		fmt.Println("[central] Warning: Cluster with even number of members has same fault tolerance, but higher quorum requirements, than cluster with one less member.")
```
fnordahl, line 41: I love the care you take to give our user a proper explanation about what is going on.

For a seasoned user, the messages do appear to be a bit long though, and they might be able to come to the conclusion of what is wrong sooner with a more terse message.

You'd be surprised how many people still operate in `80x24` terminals, and there it currently look like this:
image

Would it make sense to keep this lovely explanation for documentation and for when the client has been invoked with the `--verbose` option, and provide a terser message when `--verbose` is not provided?
https://github.com/canonical/microovn/blob/803dc03ea847d214941330c74cc0ae0027550dc0/microovn/cmd/microovn/main.go#L20

I see @mkalcok pointed out that these should go to `STDERR`, I tend to agree.

fnordahl (approved):
The points I had in the previous iteration appears to be addressed, and AFICT the ones Martin had as well. Would you agree @mkalcok ?

PR #158, opened 2024-08-01
Introduce golangci-lint and fix various issues reported by it.

fnordahl:

Reported-at: https://launchpad.net/bugs/2072471

### DIFF
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -1,0 +1,4 @@
+linters:
+  disable-all: true
+  enable:
+    - errcheck
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -2,3 +2,4 @@ linters:
   disable-all: true
   enable:
     - errcheck
+    - gosimple
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -6,3 +6,4 @@ linters:
     - govet
     - ineffassign
     - staticcheck
+    - revive
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -4,3 +4,4 @@ linters:
     - errcheck
     - gosimple
     - govet
+    - ineffassign
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -3,3 +3,4 @@ linters:
   enable:
     - errcheck
     - gosimple
+    - govet
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -5,3 +5,4 @@ linters:
     - gosimple
     - govet
     - ineffassign
+    - staticcheck
--- a/microovn/.golangci.yml
+++ b/microovn/.golangci.yml
@@ -1,3 +1,13 @@
+issues:
+  # The default exclude rules disable the requirement for GoDoc comments in the
+  # revive linter, so we can't use them.
+  exclude-use-default: false
+  # Not using the default exclude rules re-introduces the below check,
+  # disabling for now as the fact that it is excluded by default and other
+  # people are re-adding it makes it probable that we don't need it either.
+  exclude-rules:
+    - linters: [errcheck]
+      text: "Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*printf?|os\\.(Un)?Setenv). is not checked"
 linters:
   disable-all: true
   enable:
--- a/microovn/api/certificates/regenerate_ca.go
+++ b/microovn/api/certificates/regenerate_ca.go
@@ -34,7 +34,7 @@ func regenerateCaPut(s *state.State, r *http.Request) response.Response {
 		logger.Info("Re-issuing CA certificate and private key")
 		err = ovn.GenerateNewCACertificate(s)
 		if err != nil {
-			logger.Errorf("Failed to generate new CA certificate: %w", err)
+			logger.Errorf("Failed to generate new CA certificate: %v", err)
 			responseData.NewCa = false
 			return response.SyncResponse(false, &responseData)
 		} else {
@@ -44,7 +44,7 @@ func regenerateCaPut(s *state.State, r *http.Request) response.Response {
 		// Get clients for rest of the cluster members
 		cluster, err := s.Cluster(false)
 		if err != nil {
-			logger.Errorf("Failed to get a client for every cluster member: %w", err)
+			logger.Errorf("Failed to get a client for every cluster member: %v", err)
 			return response.SyncResponse(false, &responseData)
 		}
 
@@ -72,13 +72,13 @@ func regenerateCaPut(s *state.State, r *http.Request) response.Response {
 	logger.Info("Re-issuing all local OVN certificates")
 	err = ovn.DumpCA(s)
 	if err != nil {
-		logger.Errorf("%w", err)
+		logger.Errorf("%v", err)
 		return response.SyncResponse(false, &responseData)
 	}
 
 	reissuedCertificates, err := reissueAllCertificates(s)
 	if err != nil {
-		logger.Errorf("Failed to reissue certificates with new CA: %w", err)
+		logger.Errorf("Failed to reissue certificates with new CA: %v", err)
 	}
 	responseData.ReissuedCertificates[s.Name()] = *reissuedCertificates
 
--- a/microovn/api/certificates/regenerate_ca.go
+++ b/microovn/api/certificates/regenerate_ca.go
@@ -50,10 +50,10 @@ func regenerateCaPut(s *state.State, r *http.Request) response.Response {
 
 		// Bump rest of the cluster members to reissue their certificates with new CA
 		err = cluster.Query(s.Context, true, func(ctx context.Context, c *client.Client) error {
-			logger.Infof("Requesting cluster member at '%s' to re-issue its OVN certificates", c.URL())
+			clientURL := c.URL()
+			logger.Infof("Requesting cluster member at '%s' to re-issue its OVN certificates", clientURL.String())
 			result, err := microovnClient.RegenerateCA(ctx, c)
 			if err != nil {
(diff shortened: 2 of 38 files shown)


mkalcok (changes requested):
Thank you for introducing this! Static checks and linters make me happy :)

I have just one suggestion inline, otherwise LGTM. (you seem to have used wrong email address to sign the commits.)

microovn/cmd/microovnd/main.go
```diff
@@ -82,7 +80,6 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 }
 
 func init() {
```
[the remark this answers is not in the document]
fnordahl: That is a good point, I was unsure given it was not caught by the `unused` linter, but tests do appear to pass without this function, so let's get rid of it.

fnordahl:
> Thank you for introducing this! Static checks and linters make me happy :)
> 
> I have just one suggestion inline, otherwise LGTM. (you seem to have used wrong email address to sign the commits.)

Thanks! Yeah the CLA-check does not handle the Ubuntu member alias, I've raised an issue on that in canonical/has-signed-canonical-cla#47

I'm also working on enabling the `revive` linter, which among other things will give us checks for missing Godoc for public functions.

microovn/node/node.go
```diff
@@ -0,0 +1,2 @@
+// Package node provides functions operating on nodes in the cluster.
```
fnordahl, line 1: When adding package level Godoc for the `node` package, I have to admit the name appears a bit vague, having our discussion this morning in mind cc @mkalcok @crypticC0der

microovn/.golangci.yml
```diff
@@ -1,3 +1,8 @@
+issues:
+  exclude-use-default: false
+  exclude-rules:
+    - linters: [errcheck]
+      text: "Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*printf?|os\\.(Un)?Setenv). is not checked"
```
mkalcok, line 5: Could we have a short comment here explaining what's the purpose of this `exclude-rules` statement?

microovn/.golangci.yml
mkalcok, line 1: Which linter is doing the GoDoc checking? I skimmed briefly through https://golangci-lint.run/usage/linters/ but I couldn't find a mention of it.

mkalcok (approved):
Overall LGTM. I left some inline comments to consider.

microovn/api/certificates/regenerate_ca.go
```diff
@@ -34,26 +34,25 @@ func regenerateCaPut(s *state.State, r *http.Request) response.Response {
...
-		} else {
-			responseData.NewCa = true
 		}
+		responseData.NewCa = true
 
 		// Get clients for rest of the cluster members
 		cluster, err := s.Cluster(false)
 		if err != nil {
-			logger.Errorf("Failed to get a client for every cluster member: %w", err)
+			logger.Errorf("Failed to get a client for every cluster member: %v", err)
```
crypticC0der, line 46: why use %v over %w

microovn/api/ovsdb/schema_version.go
```diff
@@ -161,24 +161,28 @@ func forwardActiveSchemaVersion(s *state.State, r *http.Request, dbSpec *ovnCmd.
...
+        logger.Errorf("failed to get cluster clients: %v", err)
+		return response.ErrorResponse(500, "Internal Server Error")
+    }
 
-	for _, client_ := range clusterClients {
-		for _, node_ := range centralNodes {
-			clientUrl := client_.URL()
-			clientAddr := fmt.Sprintf("%s:%s", clientUrl.Hostname(), clientUrl.Port())
-			if clientAddr == node_.Address {
+	for _, client := range clusterClients {
```
crypticC0der, line 169: i would not use client here as its a package name

microovn/api/ovsdb/schema_version.go
```diff
@@ -161,24 +161,28 @@ func forwardActiveSchemaVersion(s *state.State, r *http.Request, dbSpec *ovnCmd.
...
+		for _, node := range centralNodes {
+			clientURL := client.URL()
+			clientAddr := fmt.Sprintf("%s:%s", clientURL.Hostname(), clientURL.Port())
+			if clientAddr == node.Address {
 				logger.Infof(
 					"Forwarding request '%s' for active %s schema to %s",
 					r.URL,
 					dbSpec.FriendlyName,
-					node_.Name,
+					node.Name,
```
crypticC0der, line 178: i would not use node as its a package name

microovn/.golangci.yml
fnordahl: Yes, it definitively deserves one, on it.

The gist of it is that the default exclude rules disables the GoDoc requirement check, so we need to disable the default exclude rules, which then also disables the one we add back.

microovn/ovn/snap.go
```diff
@@ -57,18 +57,3 @@ func snapRestart(service string) error {
 
 	return nil
 }
-
-func snapReload(service string) error {
```
crypticC0der, line 61: reasonable removal of a function :+1:

crypticC0der (dismissed):

microovn/api/ovsdb/schema_version.go
```diff
@@ -166,23 +166,23 @@ func forwardActiveSchemaVersion(s *state.State, r *http.Request, dbSpec *ovnCmd.
 		return response.ErrorResponse(500, "Internal Server Error")
     }
 
-	for _, client_ := range clusterClients {
-		for _, node_ := range centralNodes {
-			clientUrl := client_.URL()
-			clientAddr := fmt.Sprintf("%s:%s", clientUrl.Hostname(), clientUrl.Port())
-			if clientAddr == node_.Address {
+	for _, client := range clusterClients {
```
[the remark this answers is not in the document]
crypticC0der: :+1:

microovn/.golangci.yml
fnordahl: The `revive` one.

microovn/ovn/cmd/cmd.go
```diff
@@ -0,0 +1,2 @@
+// Package cmd provides functions that execute commands on the system.
```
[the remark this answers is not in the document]
fnordahl: Presumably the rendered docs would retain the fact that it is inside the `ovn` package, but spelling it out does not hurt.

microovn/api/ovsdb/schema_version.go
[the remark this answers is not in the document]
fnordahl: Ah, indeed. The linter appears to have no issue with a leading `_` as opposed to a trailing `_`, and I did not think of  any better names for the iterator variables, so I'll go with that

microovn/api/certificates/regenerate_ca.go
fnordahl: Simply because the `logger` functions do not support `%w`, which was called out by the linter.

fnordahl:
One more thing I need to fix, my editor defaults to use white space instead of tabs, so apparently I have introduced some discrepancies because of that. Will push an update shortly.

crypticC0der:
#155 merged, conflicts now exist

microovn/api/ovsdb/schema_version.go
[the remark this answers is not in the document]
crypticC0der: there are probably other names that would work here, such as clusterClient, or cClient, or vClient (v to indicate its a variable)

mkalcok:
> One more thing I need to fix, my editor defaults to use white space instead of tabs, so apparently I have introduced some discrepancies because of that. Will push an update shortly.

I'll be honest, I never knew we are using tabs in this project :D :D 

One more thing to consider before we get to merging this. I like the idea of running lint inside snap container to avoid polluting dev machine with unnecessary build dependencies. However it also means that the snap wont build if code does not pass linting, right? This could get annoying during the development, e.g. If I'm just experimenting with something new, I don't want to have to write docs for it right away.

(not a blocker, I just wanted to bring it up.)

fnordahl:
> > One more thing I need to fix, my editor defaults to use white space instead of tabs, so apparently I have introduced some discrepancies because of that. Will push an update shortly.
> 
> I'll be honest, I never knew we are using tabs in this project :D :D

Yeah, `gofmt` appears to mandate it unfortunately, otherwise I would suggest to get rid of them!

> One more thing to consider before we get to merging this. I like the idea of running lint inside snap container to avoid polluting dev machine with unnecessary build dependencies. However it also means that the snap wont build if code does not pass linting, right? This could get annoying during the development, e.g. If I'm just experimenting with something new, I don't want to have to write docs for it right away.
> 
> (not a blocker, I just wanted to bring it up.)

Yeah, this is a concern. I did not find a good answer to it though, there is no "optional" part for snapcraft AFAICT.

Perhaps we could make use of [workshop](https://github.com/canonical/workshop) in the future?

mkalcok:
> To be pragmatic while developing locally, I guess it is trivial to comment out the few lines that runs the linters as part of the build?

Yes, I suspect that this is going to be common workaround. We can get around to changing this if we find it very annoying in the future.

mkalcok:
After the merge conflict resolution, we don't have godocs for some of the exported functions ([example](https://github.com/fnordahl/microovn/blob/787e2c6371f0101e8d329413923e298374f8be14/microovn/node/node.go#L32)) but the CI doesn't seem to be catching it.

fnordahl:
> After the merge conflict resolution, we don't have godocs for some of the exported functions ([example](https://github.com/fnordahl/microovn/blob/787e2c6371f0101e8d329413923e298374f8be14/microovn/node/node.go#L32)) but the CI doesn't seem to be catching it.

Indeed, I accidentally dropped the execution of the `golangci-lint` binary during the rebase :rofl:

mkalcok (approved):
LGTM, thanks :)

PR #159, opened 2024-08-01
Use LTS microcluster

masnax:
Updates MicroOVN to use the stable release of microcluster at `v2.0.0`.

For the most part, the changes to the code in MicroOVN are just package imports and changing the definition of `state.State` to an interface. 

The actual behavioural changes are these:
* `state.Context` is no longer exposed, so `request.Context` from the API handler is passed around explicitly to each helper. This should actually result in more precise context handling for MicroOVN. 
  * One caveat is the `ovn.Refresh` hook which spawns a goroutine. For this function, a dedicated `shutdownCtx` is added to `microovnd` which will be cancelled whenever the daemon stops. 

* Adds a schema update to MicroOVN that changes the foreign key reference on the `services` table to `core_cluster_members`. This table has been renamed from `internal_cluster_members` to make it clear that accessing it is expected behaviour. I also took the liberty of renaming the previous schema update function to use an ordered naming scheme so that it doesn't become ambiguous where an update fits among the others.

* Join tokens now have expiration dates. By default, MicroOVN will set a 3 hour expiration for join tokens, but this can be configured with the `--timeout` flag added to `cluster add`.


masnax:
Test failures appear to be because of the go module transition. Should be fixed when https://github.com/canonical/microcluster/pull/219 is merged.

mkalcok:
Thanks for updating us @masnax :). Out of curiosity, is microcluster able to handle multiple schema upgrades in one upgrade step? I'm just double-checking because our current stable version of MicroOVN would, at this point, be two schema versions behind.

masnax:
Looks like the ovsdb schema upgrade test failed again, maybe because of the truststore issue.

I went ahead and added a commit that rebalances the truststore yamls as soon as the `22.03` cluster is formed, before running any tests.

mkalcok:
Did you figure out what was the issue with previous CI failings? It does seem different than what we were encountering before. Previously the issue was that one of the cluster members was showing up as "UNREACHABLE" due to the above mentioned "truststore" issue, however this time the error says
```
'Error: Database is still starting
```

masnax:
> Did you figure out what was the issue with previous CI failings? It does seem different than what we were encountering before. Previously the issue was that one of the cluster members was showing up as "UNREACHABLE" due to the above mentioned "truststore" issue, however this time the error says
> 
> ```
> 'Error: Database is still starting
> ```

I'm not sure what's causing the slowness in the test suite, but the change in behaviour during an upgrade should now be fixed.

mkalcok (approved):
LGTM and we don't seem to be hitting the upgrade issue in the CI anymore. Thanks for the update.

PR #161, opened 2024-08-07
Fix for  BUG #2074365 "microovn certificates reissue" does not check for required arguments

ismailkayi:
1.  "cmd.Help()" function will be called if no argument is provided to "microovn certificates reissue"  command. 
2. "tls_cluster_reissue_certificate_help"  test case has been added.

### DIFF
--- a/microovn/cmd/microovn/certificates_reissue.go
+++ b/microovn/cmd/microovn/certificates_reissue.go
@@ -44,7 +44,10 @@ func (c *cmdCertificatesReissue) Command() *cobra.Command {
 
 // Run method is an implementation of "microovn certificates reissue" subcommand. It requests local MicroOVN
 // service to issue new certificate for selected OVN service.
-func (c *cmdCertificatesReissue) Run(_ *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


mkalcok:
Hi @ismailkayi, thank you for the contribution. Due to the repository settings, all commits need to be [signed](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt) and [signed off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s). You don't necessarily need to sign the commit with gpg key, you can use your ssh key, you just need to configure it in your github account as a valid signing key (check out [this guide](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification). You can likely skip steps 1 and 2).

The test that you added highlights interesting, and I believe unintended, behavior. MicroOVN subcommands that fail due to the missing arguments, and return `cmd.Help()`, exit the process with return code 0. I don't think that this is right as the command essentially failed, but it's telling us otherwise. I opened [a bug](https://bugs.launchpad.net/microovn/+bug/2076246) about this that we'll look into separately.

Would you mind updating this PR and keeping only the 7090a9e commit?

PR #164, opened 2024-08-08
golangci-lint: Replace bespoke `gofmt` step with linter.

fnordahl:
Checking that `gofmt` does not have opinions is actually available as a golangci-lint linter, let's use it!

Thanks to @NucciTheBoss for pointing this one out.


mkalcok (commented):
I'm just wondering if we wat to add mention of `golang-ci` usage in `CONTRIBUTING.md` with the list of enabled checks, for better transparency. It can mention that it's run as a snap build step, perhaps also including tip on how to temporarily disable it during development process.
If some developers use IDE's that provide `golang-ci` integration, maybe it would be useful to them.

fnordahl:
> I'm just wondering if we wat to add mention of `golang-ci` usage in `CONTRIBUTING.md` with the list of enabled checks, for better transparency. It can mention that it's run as a snap build step, perhaps also including tip on how to temporarily disable it during development process. If some developers use IDE's that provide `golang-ci` integration, maybe it would be useful to them.

Describing the expectations for linting and reasons for the current implementation in `CONTRIBUTING.md` does indeed make sense.

Listing the enabled linters in the documentation does not make sense, as that is just begging for out of date/sync information. We can probably point to where we host golangci-lint's configuration though.

docs/.custom_wordlist.txt
mkalcok, line 1: Thanks for ordering these alphabetically <3

PR #165, opened 2024-08-09
Better (enabling|disabling) of central services

crypticC0der:
This changes the central service enabling to, regenerate the env file before enabling, generate certificates to allow secure communication with the other nodes.

This changes the central service disabling to properly leave the cluster and transfer ownership if they are the leader.
    
This set of patches allows central services to be successfully controlled and not cause database errors as they do in the current state of microovn (enable|disable) central

### DIFF
--- a/microovn/api/refresh.go
+++ b/microovn/api/refresh.go
@@ -1,0 +1,70 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/microcluster/v2/client"
+	"github.com/canonical/microcluster/v2/rest"
+	"github.com/canonical/microcluster/v2/state"
+
+	"github.com/canonical/microovn/microovn/api/types"
+	microovnClient "github.com/canonical/microovn/microovn/client"
+	"github.com/canonical/microovn/microovn/ovn"
+)
+
+// RegenerateEnvEndpoint defines endpoint for /1.0/env
+var RegenerateEnvEndpoint = rest.Endpoint{
+	Path: "env",
+	Post: rest.EndpointAction{Handler: regenerateEnvPost, AllowUntrusted: false, ProxyTarget: true},
+}
+
+// regenerateEnvPost implements POST method for /1.0/env endpoint.
+// This function triggers and environment update on all MicroOVN cluster members.
+// This is typically to be used with enabling and disabling entral services.
+func regenerateEnvPost(s state.State, r *http.Request) response.Response {
+	var err error
+	responseData := types.NewRegenerateEnvResponse()
+
+	// Check that this is the initial node to recive this request
+	if !client.IsNotification(r) {
+		logger.Infof("Understood notification, forwarding refresh env request")
+		// Get clients for the rest of the cluster members
+		cluster, err := s.Cluster(true)
+		if err != nil {
+			logger.Errorf("Failed to get a client for every cluster member: %v", err)
+			responseData.Success = false
+			return response.SyncResponse(false, &responseData)
+		}
+		responseData.Success = true
+
+		// Bump rest of the cluster members to regenerate their environment
+		err = cluster.Query(r.Context(), true, func(ctx context.Context, c *client.Client) error {
+			clientURL := c.URL()
+			logger.Infof("Requesting cluster member at '%s' to regenerate its environment file", clientURL.String())
+
+			_, err := microovnClient.RegenerateEnvironment(ctx, c)
+
+			if err != nil {
+				errMsg := fmt.Sprintf("Failed to contact cluster member with address %q: %s", clientURL.String(), err)
+				responseData.Errors = append(responseData.Errors, errMsg)
+			}
+			return nil
+		})
+		if err != nil {
+			return response.SmartError(err)
+		}
+	}
+
+	logger.Info("Regenerating environment file")
+
+	err = ovn.Refresh(context.Background(), r.Context(), s)
+	if err != nil {
+		logger.Errorf("Failed to refresh environment: %s", err)
+		return response.SyncResponse(false, &responseData)
+	}
+	return response.SyncResponse(true, &responseData)
+}
--- a/microovn/api/services/control.go
+++ b/microovn/api/services/control.go
@@ -33,7 +33,7 @@ func enableService(s state.State, r *http.Request) response.Response {
 		logger.Errorf("Failed to get service: %s", err)
 		return response.ErrorResponse(500, "Internal server error")
 	}
-	if !node.CheckValidService(requestedService) {
+	if !types.CheckValidService(requestedService) {
 		return response.InternalError(errors.New("Service does not exist"))
 	}
 
@@ -64,7 +64,7 @@ func disableService(s state.State, r *http.Request) response.Response {
 		logger.Errorf("Failed to get service: %s", err)
 		return response.ErrorResponse(500, "Internal server error")
 	}
-	if !node.CheckValidService(requestedService) {
+	if !types.CheckValidService(requestedService) {
 		return response.InternalError(errors.New("Service does not exist"))
 	}
 	err = node.DisableService(r.Context(), s, requestedService)
--- a/microovn/api/services/control.go
+++ b/microovn/api/services/control.go
@@ -1,6 +1,7 @@
 package services
 
 import (
+	"context"
 	"errors"
 	"net/http"
 	"net/url"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/canonical/microovn/microovn/api/types"
 	"github.com/canonical/microovn/microovn/node"
+	"github.com/canonical/microovn/microovn/ovn"
 )
 
 // ServiceControlCmd - /1.0/services/service endpoint.
@@ -36,6 +38,16 @@ func enableService(s state.State, r *http.Request) response.Response {
 	if !node.CheckValidService(requestedService) {
 		return response.InternalError(errors.New("Service does not exist"))
 	}
+	if node.SrvName(requestedService) == node.SrvCentral {
+		shutdownCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		err = ovn.Refresh(shutdownCtx, r.Context(), s)
+		if err != nil {
+			logger.Errorf("Failed to refresh environment: %s", err)
+			return response.ErrorResponse(500, "Internal server error")
+		}
+	}
+
 	err = node.EnableService(r.Context(), s, requestedService)
 	if err != nil {
 		return response.InternalError(err)
--- a/microovn/api/services/control.go
+++ b/microovn/api/services/control.go
@@ -1,7 +1,6 @@
 package services
 
 import (
-	"context"
 	"errors"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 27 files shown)


microovn/node/node.go
```diff
@@ -216,3 +221,89 @@ func ServiceWarnings(ctx context.Context, s state.State) (types.WarningSet, erro
 	}
 	return output, nil
 }
+
+// StartCentral safely starts the central service and its child service while also generating certificates
```
fnordahl, line 222: This function is doing more than starting the `central` service, which may come to a surprise to someone making use of the function in the future (including future you).

Would it make more sense to call it `JoinCentral` or something like that?

microovn/node/node.go
```diff
@@ -216,3 +221,89 @@ func ServiceWarnings(ctx context.Context, s state.State) (types.WarningSet, erro
...
+	}
+
+	err = snap.Start("ovn-northd", true)
+	if err != nil {
+		return fmt.Errorf("Failed to start OVN northd: %w", err)
+	}
+	return nil
+}
+
+// StopCentral safely stops the central service and its child service
```
fnordahl, line 255: Similar to above, this function is doing more than stopping the `central` service, which may come as a surprise to someone making use of the function in the future (including future you).

Would it make more sense to call it `LeaveCentral` or something like that?

microovn/ovn/join.go
```diff
@@ -98,33 +99,9 @@ func Join(ctx context.Context, s state.State, initConfig map[string]string) erro
 
 	// Enable OVN central (if needed).
 	if srvCentral < 3 {
```
fnordahl, line 101: Not a problem concerning this PR, but this is actually a bug. The system should probably create the cluster when there is 1 node, and hold off until there are 3 before joining two more central services to the cluster, rather than knowingly creating a cluster incapable of handling partitions when at 2 nodes.

I filed https://bugs.launchpad.net/microovn/+bug/2076436 for this.

microovn/node/node.go
```diff
@@ -216,3 +221,89 @@ func ServiceWarnings(ctx context.Context, s state.State) (types.WarningSet, erro
...
+
+	err = snap.Stop("ovn-ovsdb-server-nb", true)
+	if err != nil {
+		logger.Warnf("Failed to stop OVN NB service: %s", err)
+	}
+
+	err = snap.Stop("ovn-ovsdb-server-sb", true)
+	if err != nil {
+		logger.Warnf("Failed to stop OVN SB service: %s", err)
+	}
```
fnordahl, line 304: Now that we have higher level functions for enabling and disabling services in MicroOVN, it is also possible for the user to disable the `central` service on a node, only to come back at a later date to enable it again.

With the current code, would that actually work, or do we need to rotate out the on-disk DB files so that they are not in the way for a future (re-)join operation?

microovn/api/services/control.go
```diff
@@ -36,6 +38,16 @@ func enableService(s state.State, r *http.Request) response.Response {
 	if !node.CheckValidService(requestedService) {
 		return response.InternalError(errors.New("Service does not exist"))
 	}
+	if node.SrvName(requestedService) == node.SrvCentral {
+		shutdownCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		err = ovn.Refresh(shutdownCtx, r.Context(), s)
```
fnordahl, line 44: This appears to be a nice workaround for the fact that we currently have no mechanism to trigger cluster wide refresh on changes to the central DBs!

fnordahl (dismissed):
Overall the approach to this looks good, I do have some comments in-line.

crypticC0der:
going to add more to this pr:
- removal of central
- better re-enabling of central

microovn/node/node.go
mkalcok: This is a good point. We should probably also have functional tests that cover this scenario

tests/test_helper/bats/central_control.bats
```diff
@@ -0,0 +1,61 @@
...
+central_register_test_functions() {
+    bats_test_function \
+        --description "Testing of central node migration" \
+        -- central_tests
+    bats_test_function \
+        --description "Testing of database functionality" \
+        -- central_db_test
+}
+
+central_tests() {
```
mkalcok, line 22: Couple notes regarding the tests:
* I think that using hardcoded container names is bit fragile. You can turn the `$TEST_CONTAINERS` into an array ([example](https://github.com/canonical/microovn/blob/09213bb29ce2da8d0ae5c2d6e6163d3f56f03996/tests/test_helper/bats/ovsdb_schema_upgrade.bats#L30)) and access names via indexes
* I'd prefer checks after every step, rather than at the end
* In addition to checking microovn status and snap service status, we could also check ovsdb cluster status via `ovn-appctl`. There's already [a function that ensures that server left the cluster](https://github.com/canonical/microovn/blob/09213bb29ce2da8d0ae5c2d6e6163d3f56f03996/tests/test_helper/microovn.bash#L458-L467) but similar will probably have to be made to ensure that server joined.

microovn/api/refresh.go
```diff
@@ -0,0 +1,70 @@
...
+// regenerateEnvPost implements POST method for /1.0/env endpoint.
+// This function triggers and environment update on all MicroOVN cluster members.
+// This is typically to be used with enabling and disabling entral services.
+func regenerateEnvPost(s state.State, r *http.Request) response.Response {
+	var err error
+	responseData := types.NewRegenerateEnvResponse()
+
+	// Check that this is the initial node to recive this request
+	if !client.IsNotification(r) {
+		logger.Infof("Understood notification, forwarding")
```
mkalcok, line 34: This log message could use more context about what request is being forwarded

microovn/api/refresh.go
```diff
@@ -0,0 +1,70 @@
...
+		// Get clients for the rest of the cluster members
+		cluster, err := s.Cluster(true)
+		if err != nil {
+			logger.Errorf("Failed to get a client for every cluster member: %v", err)
+			responseData.Success = false
+			return response.SyncResponse(false, &responseData)
+		}
+		responseData.Success = true
+
+		// Bump rest of the cluster members to reissue their certificates with new CA
```
mkalcok, line 44: outdated comment

microovn/client/client.go
```diff
@@ -134,28 +134,60 @@ func getOvsdbSchemaVersion(ctx context.Context, c *client.Client, dbSpec *ovnCmd
 
 // DisableService sends request to disable service with name as specified in
 // "serviceName" argument.
-func DisableService(ctx context.Context, c *client.Client, serviceName string) (types.WarningSet, error) {
+func DisableService(ctx context.Context, c *client.Client, serviceName string) (types.WarningSet, types.RegenerateEnvResponse, error) {
```
mkalcok, line 137: Could we add the `WarningSet` to the `RegenerateEnvResponse`, to avoid returning 3 values?

mkalcok (commented):
Couple of suggestions inline, otherwise this looks sloid.

microovn/client/client.go
crypticC0der: the warning set and the regenerateEnvResponse are quite distinct and i dont personally see much against returning three values
however if we were to merge them, id move regenerateEnvResponse into warning set

mkalcok: > i dont personally see much against returning three values

It's more that I think that there's added value in returning less objects, especially if they correlate strongly. In this case you'd get single object that represents response, and contains:
* Overall success of the request
* Warnings that happened
* Errors that happened

It would make it easier for the receiver of this return value to work with it (+ you don't have to come up with two variable names to assign your function results to :laughing: ).

> however if we were to merge them, id move regenerateEnvResponse into warning set

Wouldn't the `WarningSet` be a natural subset of the overall `regenerateEnvResponse`?

microovn/api/types/services.go
```diff
@@ -88,3 +88,29 @@ func NewRegenerateEnvResponse() RegenerateEnvResponse {
 		Errors:  make([]string, 0),
 	}
 }
+
```
mkalcok, line 91: What's the reasoning for moving these vars/funcs here? I know they didn't feel super well in the `node.go` but I feel like they are very out of place in the package dedicated to API types. Did some package have problem with importing `node`?

crypticC0der: it says this in the message of the commit, it helps avoid circular dependencies, there are situations where we don't want to import node and all its dependencies just to use some types

mkalcok: Thinking about this further, the service name definitions and list of valid services indeed make sense in this package, as they are related to making valid API requests/responses. I'm still not very convinced about the `CheckValidService` being in the package reserved for API type definitions, can we keep that in the `node` package?

> it says this in the message of the commit, it helps avoid circular dependencies

It does say that. However looking this commit I don't see any circular dependencies. It seems that any packages that now imports `types` (for these newly added vars/funcs) either already imports `node` or easily could. I was interested in whether I missed something.

What do you think about moving `SrvName`, `SrvChassis`, `SrvCentral`, `SrvSwitch` to the `types` package (as proposed) and keeping `CheckValidService` in the `node`? It seems that the only place that uses `SrvChassis` right now, will have to keep importing `node` anyway.

mkalcok: nvm, let's just keep it all here. the `node` package indeed grew a lot in terms of dependencies with the addition of `Join`/`Leave` functions.

microovn/node/node.go
```diff
@@ -286,6 +287,9 @@ func leaveCentral(ctx context.Context, s state.State) error {
 		logger.Warnf("Failed to get SB database specification: %s", err)
 	}
 
+	os.Remove(paths.CentralDBNBPath())
```
fnordahl, line 290: Looking at the commit message of this commit, while I see these changes mentioned briefly at the end, I would expect only test changes and no code changes.

This is a prominent change in behavior which together with the `path.go` changes belongs in a separate commit with solid reasoning as to why this is necessary and why the chosen approach is safe.

microovn/node/node.go
fnordahl: This appears to be addressed in commit "Better (enabling|disabling) of central services" later in this series.

microovn/node/node.go
fnordahl: This appears to be addressed in commit "Better (enabling|disabling) of central services" later in this series.

microovn/node/node.go
fnordahl: This appears to be addressed by the commit currently at the tip of this series, I have however requested that commit to be split into two separate commit thus no named reference in this comment.

tests/test_helper/microovn.bash
```diff
@@ -501,6 +501,52 @@ function wait_ovsdb_cluster_container_leave() {
     return $rc
 }
 
+# wait_ovsdb_cluster_container_join SERVER_ID CONTROL_PATH DB_NAME TIMEOUT CONTAINER1 [CONTAINER2 ...]
```
fnordahl, line 504: nit: there is a generic `wait_until` function that could be used for the wait part of this, but I won't block on this. 

Example usage:
https://github.com/canonical/microovn/pull/138/commits/d3fae453133427edeb8ff6fd6b9b568391140dee#diff-99adb05bce6de38bad928c437a3672fff44ed684f06fff6f820edcae6954d3c3R114-R116

fnordahl (changes requested):

fnordahl (approved):
LGTM, thanks!

PR #169, opened 2024-08-13
regenerate_ca: Fix requests not being forwarded

mkalcok:
API endpoint for regenerating CA certificates was broken after recent upgrade to `microcluster` library[0] that changed signature of the `State.Cluster()` method. This method now accepts boolean value that must be set to `true` if we want recipient to properly recognized the request with the `client.IsNotification()` method.

Current code was not properly recognizing the initial request from the client and as a consequence, it did not trigger re-issue of the CA certificate and subsequent notification to the rest of the cluster.

[0] https://github.com/canonical/microovn/commit/5b42b2ed7a291de6c5a5551b5b3edfbb35a45adb

### DIFF
--- a/microovn/api/certificates/regenerate_ca.go
+++ b/microovn/api/certificates/regenerate_ca.go
@@ -29,7 +29,7 @@ func regenerateCaPut(s state.State, r *http.Request) response.Response {
 	responseData := types.NewRegenerateCaResponse()
 
 	// Check that this is the initial node that received the request and recreate new CA certificate
-	if client.IsNotification(r) {
+	if !client.IsNotification(r) {
 		// Only one recipient of this request needs to generate new CA
 		logger.Info("Re-issuing CA certificate and private key")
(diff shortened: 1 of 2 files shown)


microovn/api/certificates/regenerate_ca.go
```diff
@@ -29,7 +29,7 @@ func regenerateCaPut(s state.State, r *http.Request) response.Response {
 	responseData := types.NewRegenerateCaResponse()
 
 	// Check that this is the initial node that received the request and recreate new CA certificate
-	if client.IsNotification(r) {
+	if !client.IsNotification(r) {
```
fnordahl, line 32: Wow, 5b42b2ed7a291de6c5a5551b5b3edfbb35a45adb even inverted this check when transitioning from `IsForwardedRequest` to `IsNotification`, which is not great.

fnordahl (approved):
LGTM, thank you for fixing this and for the increased test coverage!

PR #171, opened 2024-08-15
tests: wait before upgrading cluster

mkalcok:
Occasionally we encounter failure in `ovsdb_schema_upgrade` test related to the internal Dqlite schema upgrade. When it happens, one of the cluster members appears "UNREACHABLE" for extended amount of time.
It was suggested that this occurs when the test triggers upgrade before the original cluster had chance to converge. This change adds check that waits for all cluster members to appear "ONLINE" before it continues with upgrade.


fnordahl (approved):
LGTM, thanks!

PR #172, opened 2024-08-15
snap: Pin go to 1.22/stable.

fnordahl:
The `go` build-snap was recently updated to 1.23 which for some reason makes `golangci-lint` go OOM.

Until we figure out the root of the issue let's get the gate unblocked.


mkalcok (dismissed):
LGTM thanks for tracking this down

crypticC0der (approved):
LGTM :sunglasses: :+1:

PR #174, opened 2024-08-19
Update to developer documentation on LXD remote support and various test helper updates.

fnordahl:

These changes are made for #138 but are also useful to others, so let's extract them to separate PR while awaiting completion of said PR.

### DIFF
--- a/tests/test_helper/common.bash
+++ b/tests/test_helper/common.bash
@@ -218,3 +218,52 @@ function netns_ifadd() {
     lxc_exec "$container" \
         "ip netns exec $name ip link set up dev $ifname"
 }
+
+function _netns_dst_base_name() {
+    local dst=$1; shift
+    local netns=$1
+
+    echo "${netns:+$netns-}${dst//[\.:]/_}"
+}
+
+# ping_start CONTAINER DST [ NETNS ]
+#
+# Start ping to DST in CONTAINER in the background, optionally using network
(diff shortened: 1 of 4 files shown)


tests/test_helper/common.bash
```diff
@@ -218,3 +218,42 @@ function netns_ifadd() {
...
+#
+# Output from ping and the PID will be recorded as files in '/tmp/' in
+# CONTAINER, and the process and results can be reaped by a subsequent call to
+# ``ping_reap``.
+function ping_start() {
+    local container=$1; shift
+    local dst=$1; shift
+    local netns=$1
+
+    local base_filename="${netns:+$netns-}${dst//[\.:]/_}"
```
mkalcok, line 235: I wonder if it would be worth to turn the filename creation into a one-line function that both `ping_start` and `ping_reap` call. To keep it consistent.

mkalcok (changes requested):
One small in-line request. Otherwise LGTM. Thanks for these helpers, they'll come in very handy for BGP tests as well.

tests/test_helper/common.bash
fnordahl: Yea, it's a lot of variable passing and echoing back though, which I think is why I opted not to.

But you're right consistency and correctness trumps that

mkalcok (approved):
LGTM, thanks

PR #175, opened 2024-08-21
Removal of deprecated central snap service

crypticC0der:
This removes the microovn.central service. This has been coming a long time and removing it should ease confusion between the central service and the central snap service.

This service has been discouraged for a long time and doesn't handle its "child" services particularly well and people shouldn't really be interacting with these snap services anyway as we have enablement and disablement of services within the codebase


fnordahl (commented):
It is indeed time to get rid of the transitional `snap.microovn.central` service!

From cursory view, I see mentions of `microovn.central` in `docs/how-to/tls.rst`.

For now I'd propose you replace those with the underlying system service names, and we can add a bug for revisiting that section to either archive it or rework it to use the higher level service functionality.

What do you think?

fnordahl (approved):
LGTM, thanks!

PR #176, opened 2024-08-21
Newdocs

crypticC0der:
This changes the current service documentation to be about how microovn
handles services, and the service available through enable/disable.

This also includes a howto documentation going over an example of
service control using microovn enable/disable

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -6,7 +6,9 @@ LXD
 Livepatch
 Makefile
 MicroOVN
+MicroOVN's
 OVN
+OVN's
 OVS
 OpenStack
 PKI
--- a/docs/reference/services.rst
+++ b/docs/reference/services.rst
@@ -4,9 +4,60 @@
 MicroOVN services
 =================
 
+MicroOVN functionality is separated into distinct services that can be easily
+controlled via ``microovn enable`` and ``microovn disable``.
+
 This page presents a list of all MicroOVN services. Their descriptions are
 for reference only - the user is not expected to interact directly with these
 services.
+
+Handling services with enable/disable
+-------------------------------------
+
+The status of all services is displayed by running:
+
+.. code-block:: none
+
+   microovn status
+
+``central service``
+-------------------
+
+This is responsible for the database control. The database is clustered and uses
+the `RAFT <https://docs.openvswitch.org/en/latest/ref/ovsdb.7/#clustered-database-service-model>`_
+algorithm for consensus it can handle (n-1)/2 failures, where n is the number of
+nodes.
+
+Central is enabled on a new node whenever there are less than 3 nodes running
+the central services
+
+This service controls the following `Snap services`_:
+
+- ``microovn.ovn-ovsdb-server-nb``
+- ``microovn.ovn-ovsdb-server-sb``
+- ``microovn.ovn-northd``
+
+``chassis service``
+-------------------
+
+This service controls the ``ovn-controller`` daemon, which is OVN's agent on each
+hypervisor and software gateway. It is a distributed component running on the
+side of every Open vSwitch instance.
+It is enabled by default.
+
+The snap service this controls is ``microovn.chassis``
+
+`switch service``
+-------------------
+
+This service ``Open vSwitch`` and ensures its running properly. Much like chassis it
+is enabled by default.
+
+The snap service this controls is ``microovn.switch``
+
+
+Snap services
+-------------
 
 The status of all services is displayed by running:
 
--- a/docs/how-to/index.rst
+++ b/docs/how-to/index.rst
@@ -14,3 +14,4 @@ a basic understanding of OVN.
    logs
    major-upgrades
    ovn-underlay
+   service-control
--- a/docs/how-to/service-control.rst
+++ b/docs/how-to/service-control.rst
@@ -1,0 +1,157 @@
+===============
+Service Control
+===============
+
+Service control refers to the set of commands for enabling and disabling a
+given MicroOVN service. MicroOVN has a set of services referred to here
+:doc:`Services Reference </reference/services>`, which are responsible for
+handling core functionality.
+
+You can disable services manually using snap, but the service control does not
+update the desired state or handle joining clusters and configuring the service
+properly, hence the strong reasoning to interact with services through this
+method.
+
+.. note::
+
+   This assumes you have MicroOVN installed and a clustered across three of more
+   nodes. These nodes will be referred to as ``first``, ``second`` and ``third``
+   respectively.
+
+Disabling a MicroOVN service
+----------------------------
+
+Disabling a MicroOVN service will configure it to not start automatically at
+boot and stop the service if it is running.
+
+run on ``first``:
+
+.. code-block:: none
+
+   microovn disable switch
+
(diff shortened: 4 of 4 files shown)


docs/how-to/migration.rst
```diff
@@ -0,0 +1,118 @@
...
+run on ``first``:
+
+.. code-block:: none
+
+   microovn enable switch
+   Service switch enabled
+
+.. warning::
+
+   If the switch service is enabled you may get an error, this is fine.
```
fnordahl, line 79: I'd suggest to drop this warning, if you *really* want to keep it, downgrade it to a `note` and phrase it more generically, such as "If the service is already enabled you will get an error message."

docs/how-to/migration.rst
```diff
@@ -0,0 +1,118 @@
...
+.. code-block:: none
+
+   snap services microovn | grep switch
+   microovn.switch                  enabled     active  -
+
+You should be able to see here that the service is running and enabled on
+startup. The other nodes are also aware of this as if you query the status you
+will see it there and running.
+
+run on ``second``:
```
fnordahl, line 97: As noted above, let's drop mention of the snap / systemd services on this page.

docs/how-to/migration.rst
```diff
@@ -0,0 +1,118 @@
...
+
+.. code-block:: none
+
+   microovn status
+   MicroOVN deployment summary:
```
fnordahl, line 102: Let's keep the command and their output in separate code blocks, there is an example of this from the MicroOVN documentation set here: https://canonical-microovn.readthedocs-hosted.com/en/latest/how-to/tls/#common-issues

docs/how-to/migration.rst
```diff
@@ -0,0 +1,118 @@
...
+   OVN Database summary:
+   OVN Northbound: OK (7.3.0)
+   OVN Southbound: OK (20.33.0)
+
+Uses
+----
+
+Typically the most common use case of this will be to control the nodes the
+central services are running on and to increase the number of central services
+beyond the default of 3. Though this will also be used to enable BGP routing.
```
fnordahl, line 118: While the last sentence is true, let's add it when that actually lands.

docs/.custom_wordlist.txt
```diff
@@ -11,6 +11,8 @@ yaml
 RTD
 readthedocs
 MicroOVN
+MicroOVN's
+BGP
```
fnordahl, line 15: Let's not add this before the feature actually lands.

docs/.custom_wordlist.txt
```diff
@@ -47,4 +49,4 @@ devel
 YY
 hostname
 Geneve
-Virtualization
\ No newline at end of file
+Virtualization
```
fnordahl, line 52: Thank you for restoring newlines at the end of the file!

We should add a check for this in the `check-lint` step:
https://github.com/canonical/microovn/blob/4a25d8ef273bb4de70d6a7632920738d3bfeea8e/Makefile#L16-L21

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
 MicroOVN services
 =================
 
+MicroOVN has lots of its functionality separated into distinct services that can
```
fnordahl, line 7: Drop "has lots of its", `MicroOVN's functionality is ...`

A slight digression no action requested now, our use of the word `service` for multiple concepts is a bit unfortunate, I guess this why sunbeam went with a role concept, but that is for another day.

Let's split the entire page into two sections, leaving the original content intact, apart from changing `page` to `section`.

Like this:
```rst
=================
MicroOVN services
=================

*Your new content*.

snap services
-------------
This section ...
```

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+be controlled via MicroOVN enable/disable or snap/systemd, only the former is
```
fnordahl, line 8: exemplify the commands, i.e. `controlled via ``microovn enable`` and ``microovn disable`` commands.

Drop mentioning snap/systemd and the comment after the comma.

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+The status of all services is displayed by running:
+
+.. code-block:: none
+
+   microovn status
+
+``central service``
+-------------------
+
+This is responsible for the database control. The database is a RAFT database so
```
fnordahl, line 27: ... database is clustered and uses the Raft algorithm for consensus.  It can ...

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+it can handle (n-1)/2 failures, where n is the number of nodes.
```
fnordahl, line 28: Would it make sense to add a link to the [OVSDB clustered database service model](https://docs.openvswitch.org/en/latest/ref/ovsdb.7/#clustered-database-service-model) documentation?

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+
+Central is enabled on a new node whenever there are less than 3 nodes running
+the central services
+
+This service controls the following snap services:
```
fnordahl, line 33: Add page-internal link to the `snap services` heading.

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+
+- ``microovn.ovn-ovsdb-server-nb``
+- ``microovn.ovn-ovsdb-server-sb``
+- ``microovn.ovn-northd``
+
+``chassis service``
+-------------------
+
+This service controls the OVN control and MicroOVN's interaction with OVN.
```
fnordahl, line 42: Did you mean `ovn-controller` daemon?

What is meant by `MicroOVN's interaction with OVN` ?

For the uninitiated, the use of the word `controller` may be confusing, so it is worth adding a sentence explaining that it is a distributed component running on the side of every Open vSwitch instance. Better examples of wording may be found on the [OVN architecture page](https://manpages.ubuntu.com/manpages/oracular/en/man7/ovn-architecture.7.html)

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+It is enabled by default.
+
+The snap service this controls is ``microovn.chassis``
+
+`switch service``
+-------------------
+
+This service ``OpenVSwitch`` and ensures its running properly. Much like chassis it
```
fnordahl, line 50: Open vSwitch

docs/reference/services.rst
```diff
@@ -4,10 +4,64 @@
...
+The snap service this controls is ``microovn.switch``
+
+
+Handling services with snap
+----------------------------
+
+A user is not expected nor advised to interact with services via snap or
+systemd, the recommended way to do this is the MicroOVN service control
+mentioned above. Controlling services through snap/systemd may lead to services
+not working correctly if at all.
```
fnordahl, line 62: Drop this section, the explanation at the top of the page is sufficient.

fnordahl (changes requested):
The PR needs to be rebased as the first commit is a different version of what is already is committed to the main branch.

fnordahl:
@crypticC0der looking forward to an update on this PR, would be good to get it in now that the feature has merged

fnordahl (approved):
LGTM, thanks!

PR #177, opened 2024-08-29
microovn: Use OVN major version in DaemonArgs.

fnordahl:
At present we provide a git hash as the version in DaemonArgs. Other microcloud daemons use this information to determine compatibility, so let's use our main component major version instead.


microovn/version/version_test.go
```diff
@@ -0,0 +1,32 @@
+package version
+
+import "testing"
```
crypticC0der, line 3: ah unit tests, lovely, we needed some of them

microovn/cmd/microovnd/main.go
```diff
@@ -14,11 +14,9 @@ import (
 	"github.com/canonical/microovn/microovn/api"
 	"github.com/canonical/microovn/microovn/database"
 	"github.com/canonical/microovn/microovn/ovn"
+	"github.com/canonical/microovn/microovn/version"
 )
 
-// MicroOvnVersion - the version string for the MicroOVN MicroCluster daemon.
-var MicroOvnVersion string
```
crypticC0der, line 20: yes good it didnt make much sense for versions to be here and in microovn/main

crypticC0der (approved):

PR #180, opened 2024-09-04
Give ovndb commands timeouts and multiple attempts

crypticC0der:
Reported-at: https://bugs.launchpad.net/microovn/+bug/2077573

This aims to fix the issue with commands hanging and therby blocking, this adds a timeout to the command and the command will be attempted at most 3 times.

This should solve the reported bug

### DIFF
--- a/microovn/ovn/cmd/commands.go
+++ b/microovn/ovn/cmd/commands.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"strconv"
 
 	"github.com/canonical/lxd/shared"
@@ -106,11 +107,14 @@ func WaitForDBState(ctx context.Context, _ state.State, db *OvsdbSpec, dbState s
 	return nil
 }
 
-// ovnDBCtl is a helper function to execute "ovn-nbctl" and "ovn-sbctl" commands. It takes "dbType" parameter
-// that identifies which database server it's going to talk to and "args" parameters which is list of
-// arguments that are directly passed to ovn-nbctl/ovn-sbctl commands. Before the command is executed, this
-// function ensures that underlying database is in connected state. If the database does not reach "connected"
-// state before specified timeout, an error is returned.
+// ovnDBCtl is a helper function to execute "ovn-nbctl" and "ovn-sbctl" commands
+// which are re-tried up to 3 times. If command arguments do not specify timeout (-t or
+// --timeout), a default of 30s will be added automatically. It takes "dbType"
+// parameter that identifies which database server it's going to talk to and
+// "args" parameters which is list of arguments that are directly passed to
+// ovn-nbctl/ovn-sbctl commands. Before the command is executed, this function
+// ensures that underlying database is in connected state. If the database does
+// not reach "connected" state before specified timeout, an error is returned.
 func ovnDBCtl(ctx context.Context, s state.State, dbType OvsdbType, timeout int, args ...string) (string, error) {
 	var baseCmd string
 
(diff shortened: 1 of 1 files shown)


microovn/ovn/cmd/commands.go
```diff
@@ -132,7 +132,14 @@ func ovnDBCtl(ctx context.Context, s state.State, dbType OvsdbType, timeout int,
...
 
-	return shared.RunCommandContext(ctx, baseCmd, args...)
+	// try command 3 times if it is failing
+	for attempts := 0; attempts < 3; attempts++ {
+		output, err := shared.RunCommandContext(ctx, baseCmd, append(args, "--timeout", "30")...)
+		if err == nil {
+			return output, nil
+		}
+	}
+	return "", err
```
mkalcok, line 142: +1 for timeout here. I'm just wondering if it's necessary for us to retry it 3 times. OVS/OVN commands are pretty good at figuring out when retry helps and when not. For example if connection fails, `ovn-nbctl` is going to retry with increasing backoff. However if there's something like wrong argument, it will hard fail without retries. If our intention is to have 90s timeout, we can just put it in there directly, and let ovn figure out when retries are helpful and when not.

Note: there's a `VSCtl` function that executes `ovs-vsctl` and that might benefit from the `--timeout` argument as well.

microovn/ovn/cmd/commands.go
```diff
@@ -167,7 +183,14 @@ func VSCtl(ctx context.Context, s state.State, args ...string) (string, error) {
 		return "", err
 	}
 
-	return shared.RunCommandContext(ctx, "ovs-vsctl", args...)
+	// try command 3 times if it is failing
+	for attempts := 0; attempts < 3; attempts++ {
+		output, err := shared.RunCommandContext(ctx, "ovs-vsctl", append(args, "--timeout", "30")...)
```
mkalcok, line 188: I suspect that the test failures are because the `--timeout 30` is added at the end of the command. This may be messing up the expectations of the `ovs-vsctl`.
I don't think there's a `prepend` functions for slices in GO, so you might need to do something like this https://github.com/canonical/microovn/blob/96c619d8e3173ecb338bca1bd059490bdf6fa96b/microovn/ovn/cmd/commands.go#L176-L183

microovn/ovn/cmd/commands.go
```diff
@@ -167,7 +188,18 @@ func VSCtl(ctx context.Context, s state.State, args ...string) (string, error) {
 		return "", err
 	}
 
-	return shared.RunCommandContext(ctx, "ovs-vsctl", args...)
+	if !slices.Contains(args, "--timeout") {
```
mkalcok, line 191: `-t` is also a valid way to specify timeout.

microovn/ovn/cmd/commands.go
```diff
@@ -106,11 +107,13 @@ func WaitForDBState(ctx context.Context, _ state.State, db *OvsdbSpec, dbState s
 	return nil
 }
 
-// ovnDBCtl is a helper function to execute "ovn-nbctl" and "ovn-sbctl" commands. It takes "dbType" parameter
-// that identifies which database server it's going to talk to and "args" parameters which is list of
-// arguments that are directly passed to ovn-nbctl/ovn-sbctl commands. Before the command is executed, this
-// function ensures that underlying database is in connected state. If the database does not reach "connected"
-// state before specified timeout, an error is returned.
+// ovnDBCtl is a helper function to execute "ovn-nbctl" and "ovn-sbctl" commands
+// which are tried up to 3 times with a 30s timeout.. It takes "dbType"
```
mkalcok, line 111: We should mention that the `30s` timeout value is just a fallback in case user didn't specify timeout in CLI args.

mkalcok (changes requested):
Couple of nits inline, but otherwise LGTM

mkalcok (approved):
LGTM, thanks!

PR #181, opened 2024-09-12
Make err global scope for command retries

crypticC0der:
due to a bug in a previous commit the err variable was local and therin if the command failed three times, it would not be returned due to the err in the loop being a local variable declared by the walrus operator


microovn/ovn/cmd/commands.go
```diff
@@ -142,6 +142,7 @@ func ovnDBCtl(ctx context.Context, s state.State, dbType OvsdbType, timeout int,
 
 	// try command 3 times if it is failing
 	for attempts := 0; attempts < 3; attempts++ {
+		var output string
 		output, err := shared.RunCommandContext(ctx, baseCmd, args...)
```
mkalcok, line 146: The `:=` needs to be changed to `=` here as well.

mkalcok (changes requested):
Thanks for the quick fix. There's still one shorthand variable declaration that needs to be fixed, otherwise LGTM.

mkalcok (approved):
LGTM, thanks!

PR #183, opened 2024-09-12
Attempt to fix services restart test

crypticC0der:
The services restart test is flaky at the moment, This is due to elections needing to take place within the central databases for leaves and joins and this on low end hardware like the github CI are slow.

This should wait until the containers are online before continuing

### DIFF
--- a/microovn/cmd/microovn/cluster_bootstrap.go
+++ b/microovn/cmd/microovn/cluster_bootstrap.go
@@ -43,7 +43,7 @@ func (c *cmdClusterBootstrap) Run(cmd *cobra.Command, args []string) error {
 
 	// Get system address.
 	address := util.NetworkInterfaceAddress()
-	address = util.CanonicalNetworkAddress(address, 6443)
+	address = util.CanonicalNetworkAddress(address, DefaultMicroClusterPort)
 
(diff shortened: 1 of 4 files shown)


mkalcok (changes requested):
`wait_microovn_online` checks status of the internal Dqlite database and whether all members are "ONLINE". None of the enable/disable commands executed here change that state, so it doesn't make much sense to wait for ONLINE here. This likely works/helps just because it adds some execution time, but we should come up with more targeted "waits", if they are necessary.

crypticC0der:
> `wait_microovn_online` checks status of the internal Dqlite database and whether all members are "ONLINE". None of the enable/disable commands executed here change that state, so it doesn't make much sense to wait for ONLINE here. This likely works/helps just because it adds some execution time, but we should come up with more targeted "waits", if they are necessary.

this is true, I do think a simple sleep of 10 seconds (or 3x the database election time) would fix it, but im aware this may be bad practice

mkalcok:
Yes, unconditional sleeps are a slippery slope. 10 seconds here, 5 seconds there, and suddenly the tests spend 2-3 minutes just sleeping.
Perhaps you could use something like
```
ovsdb-client wait <path/to/db.sock> <DB_NAME> connected
```

PR #185, opened 2024-09-30
Revert "Coverage"

mkalcok:
We did not consider that to publish our snap via launchpad, we require static `snapcraft.yaml` file present in the repository.

Reverts canonical/microovn#184


mkalcok:
Superseded by #186 because PR's created via web UI are not properly signed.

PR #187, opened 2024-10-01
Coverage - 2nd attempt

mkalcok:
after #184 had to be reverted, this is a second attempt at enabling coverage data collection during system tests.

This time the `snapcraft.yaml` file is static, so build system like `launchpad` shouldn't have problem using this repository to publish snaps. Instead, the build is controlled via variables in `microovn/.build_control` file that gets sourced during the snap build.

This PR is a draft. It does not contain doc changes yet as I'd like some feedback on its viability first.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,7 +59,10 @@ jobs:
               --classic
 
       - name: Build snap
-        run: make $MICROOVN_SNAP
+        run: |
+          # Build snap with coverage support
+          sed -i 's/MICROOVN_COVERAGE=.*/MICROOVN_COVERAGE="yes"/g' microovn/build-aux/environment
+          make $MICROOVN_SNAP
 
       - name: Upload artifacts
         if: always()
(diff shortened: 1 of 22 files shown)


fnordahl (approved):

fnordahl:
> after #184 had to be reverted, this is a second attempt at enabling coverage data collection during system tests.
> 
> This time the `snapcraft.yaml` file is static, so build system like `launchpad` shouldn't have problem using this repository to publish snaps. Instead, the build is controlled via variables in `microovn/.build_control` file that gets sourced during the snap build.
> 
> This PR is a draft. It does not contain doc changes yet as I'd like some feedback on its viability first.

Yes, I think it's a viable path, a couple of comments:

1.
The file augments the snap build environment, should we perhaps keep the file together with `snapcraft.yaml`? I see no need to tuck it away as a hidden file btw.

2.
Should we augment the version string to make it clear this is a instrumented snap/binary, in case someone manages to toggle this unintentionally.

mkalcok:
> The file augments the snap build environment, should we perhaps keep the file together with snapcraft.yaml? 

The main benefit of having the file in `microovn/` directory is that if its content changes, snapcraft will pick it up on it and rebuild the `microovn` part. If it was in the `snap/` and you changed it, you'd just get `Skipping build for microovn (already ran)`

> Should we augment the version string to make it clear this is a instrumented snap/binary

So for the snap version, we are already cutting it pretty close to the character limit of 32 (e.g. `24.03.2+snap9c47bdc598-dirty` is 28 chars), but we could include it in the version string that gets outputed by `microovn --version`

fnordahl:
> > The file augments the snap build environment, should we perhaps keep the file together with snapcraft.yaml?
> 
> The main benefit of having the file in `microovn/` directory is that if its content changes, snapcraft will pick it up on it and rebuild the `microovn` part. If it was in the `snap/` and you changed it, you'd just get `Skipping build for microovn (already ran)`

That is a fair point, what about `microovn/build-aux/environment` or something like that then?

> > Should we augment the version string to make it clear this is a instrumented snap/binary
> 
> So for the snap version, we are already cutting it pretty close to the character limit of 32 (e.g. `24.03.2+snap9c47bdc598-dirty` is 28 chars), but we could include it in the version string that gets outputed by `microovn --version`

The snap version is indeed long as-is, if we can watermark the binary that works for me.

mkalcok:
This approach seems to be working fine. CI results produced `cobertura.xml` that looks good.

Update since last push:
* control file moved to `microovn/build-aux/environment`
* `microovn --version` shows `-cover` suffix if binaries were built with coverage instrumentation

fnordahl (approved):
LGTM, thanks!

PR #188, opened 2024-10-04
refactor: centralize OVS/OVN_RUNDIR usage [lp:2042843]

wideawakening:
addressing some small bites as part of the onboarding, suggested by @mkalcok 
https://bugs.launchpad.net/microovn/+bug/2042843

- [ ] check runtime `$SNAP_COMMON/data/env/ovn.env` vs.  constant/unchanging `$SNAP/ovn.env` 
i'm explicitly putting the runtime  config after the static one in case there is an override
  ```
  . "${SNAP}/ovn.env"
  . "${SNAP_COMMON}/data/env/ovn.env"
  ````

side-quest
- [ ] if you see value on it, i can try find a way to help maintain the export proliferation on keys that exist on the `ovn.env` file

sorry for the extra commits while testing sign-off and signature 🙏🏽

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,7 +59,10 @@ jobs:
               --classic
 
       - name: Build snap
-        run: make $MICROOVN_SNAP
+        run: |
+          # Build snap with coverage support
+          sed -i 's/MICROOVN_COVERAGE=.*/MICROOVN_COVERAGE="yes"/g' microovn/build-aux/environment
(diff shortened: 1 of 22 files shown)


wideawakening:
continued on #189
---

this was unexpected. i just wanted to flatten/cleanup the signoff/signature tests, since the action checks on all commits and its on failed status.

force pushed to restore, and seems it closed the fork 😓  sorry for that. not sure howto do that for the next time it happens, happy to take a suggestion.

PR #191, opened 2024-10-18
docs: Add cryptographic reference documentation.

fnordahl:
It would be useful for our users to be able to look up the details of how MicroOVN makes use of cryptography in its implementation.

### DIFF
--- a/docs/reference/cryptography.rst
+++ b/docs/reference/cryptography.rst
@@ -1,0 +1,29 @@
+============
+Cryptography
+============
+
+Transport layer security (TLS)
+------------------------------
+All network endpoints exposed by MicroOVN services are secured using multiple
+components of the `TLS protocol`_, including encryption, authentication and
+integrity.  Through the use of the `Ubuntu OpenSSL`_ packages, TLS versions
+below 1.2 are disabled for security reasons.
+
+There are two self-signed certificate authorities in use, one for the
+`MicroCluster`_ based ``microovnd`` daemon, another for the `OVN`_ daemons.
+These are initialised during the initial bootstrap of the cluster.
+
+Keys are generated using a 384 bit `Elliptic Curve`_ algorithm often referred
+to as P-384.
+
+Both sets of daemons are by default configured to make use of TLS to encrypt
+on the wire communication, as well as using certificate data for authenticating
+and verifying remote peers, ensuring only trusted components can participate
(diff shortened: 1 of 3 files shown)


docs/reference/cryptography.rst
```diff
@@ -0,0 +1,28 @@
+============
+Cryptography
+============
+
+Transport Layer Security (TLS)
```
crypticC0der, line 5: Shouldnt we be using sentence case for headings?

docs/reference/cryptography.rst
```diff
@@ -0,0 +1,28 @@
...
+All network endpoints exposed by MicroOVN services are secured using multiple
+components of the `TLS protocol`_, including encryption, authentication and
+integrity.  Through the use of the `Ubuntu OpenSSL`_ packages, TLS versions
+below 1.2 are disabled for security reasons.
+
+There are two self-signed certificate authorities in use, one for the
+`MicroCluster`_ based ``microovnd`` daemon, another for the `OVN`_ daemons.
+These are initialised during the initial bootstrap of the cluster.
+
+Keys are generated using a 384 bit Elliptic Curve algorithm often referred to
```
crypticC0der, line 16: would it be worth linking to further information on this algorithm somewhere?

crypticC0der (commented):
Really really good docs
I have some very minor notes, but LGTM

docs/reference/cryptography.rst
fnordahl: Ah, indeed, thank you for pointing that out!

docs/reference/cryptography.rst
fnordahl: Yeah, we probably should. At first attempt the best reference I found looked a bit outdated, which made me unsure, I think we need something better than Wikipedia.

I'll make another attempt at finding a more recent reference.

fnordahl: It's at least referred to in RFC 5480, let's try that as a reference.

fnordahl: @crypticC0der pointed to the wikipedia page on [elliptic curve cryptography](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography), while generic it is quite decent, let's use it!

crypticC0der (approved):
LGTM!

PR #192, opened 2024-10-22
[branch-22.03] tests: fix missing wait in TLS test suite

mkalcok:
TLS test suite setup was missing call to `wait_containers_ready` that other test suites utilize. It sometimes caused the tests to error out with:

error: too early for operation, device not yet seeded or device model not acknowledged

(cherry picked from commit bc8f87b1c69399c732a55e9bfb171839a96312ad)

Alongside the backport, we also increased timeout for OVN DB convergence after snap upgrade.

### DIFF
--- a/tests/test_helper/bats/upgrade.bats
+++ b/tests/test_helper/bats/upgrade.bats
@@ -25,8 +25,8 @@ setup() {
         if [[ "$container_services" != *"central"* ]]; then
             continue
         fi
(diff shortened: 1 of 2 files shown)


fnordahl (approved):
The first commit is a cherry-pick, the second commit is not but that is fine because this specific functionality has moved on the main branch.

LGTM, thanks!

PR #194, opened 2024-10-24
Add defaults for initial and and connect addresses

crypticC0der:
Previously if the enviroment file was generated before any central
services had been enabled, then the initial and connect fields would be
empty, stopping northd from connecting to south and north databases.

This fix gives these defaults, which will only be used in the case of a
no central being found, in which case a central will always be generated
on the local machine, so the default addresses will be accurate.

### DIFF
--- a/tests/test_helper/bats/cluster.bats
+++ b/tests/test_helper/bats/cluster.bats
@@ -168,7 +168,7 @@ cluster_ovs_ovn_remote_addresses() {
     done
 }
 
-# _test_db_connection_string NBSB
+# cluster_test_db_connection_string NBSB
 #
 # Tests that database connection string for NBSB contains the expected
 # addresses.
@@ -211,7 +211,7 @@ cluster_test_db_connection_string() {
     done
 }
 
-# _test_southbound_propagation
+# cluster_test_southbound_propagation
 #
 # Tests that database connection between northbound and southbound databases
 # is properly functional.
--- a/tests/test_helper/bats/cluster.bats
+++ b/tests/test_helper/bats/cluster.bats
@@ -35,6 +35,9 @@ cluster_register_test_functions() {
     bats_test_function \
         --description "Open_vSwitch external_ids:ovn-remote addresses" \
         -- cluster_ovs_ovn_remote_addresses
+    bats_test_function \
+        --description "Ensure northd connectivity between NB and SB" \
+        -- cluster_test_southbound_propagation
 }
 
 cluster_expected_count() {
@@ -205,6 +208,25 @@ cluster_test_db_connection_string() {
             # partial matching.
             assert_output -p "ssl:${expected_addr}:${expected_port}"
         done
+    done
+}
+
+# _test_southbound_propagation
+#
+# Tests that database connection between northbound and southbound databases
+# is properly functional.
+#
+# The command tested is one which waits for the change to propagate to the
+# southbound database and was specficially picked as it was at one point
+# failing until pointed out by microcloud
+cluster_test_southbound_propagation() {
+    for container in $TEST_CONTAINERS; do
+        run lxc_exec "$container" \
+            "microovn.ovn-nbctl --timeout=10 --wait=sb ha-chassis-group-add testnet"
+        assert_success
+        run lxc_exec "$container" \
+            "microovn.ovn-nbctl --timeout=10 --wait=sb ha-chassis-group-del testnet"
+        assert_success
     done
 }
 
--- a/microovn/ovn/environment.go
+++ b/microovn/ovn/environment.go
@@ -122,6 +122,22 @@ func generateEnvironment(ctx context.Context, s state.State) error {
 		localAddr = "[" + localAddr + "]"
 	}
 
+	//set northbound to be at the local address if there is no central db found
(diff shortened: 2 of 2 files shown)


microovn/ovn/bootstrap.go
```diff
@@ -59,15 +59,15 @@ func Bootstrap(ctx context.Context, s state.State, initConfig map[string]string)
 		return err
 	}
 
-	err = node.EnableService(ctx, s, types.SrvCentral)
+	err = generateEnvironment(ctx, s)
```
mkalcok, line 62: I don't think I quite understand. Why does this call to `generateEnvironment` fixes this issue, when there's call to this same function 20 lines above? How does enabling `switch` makes a difference?

crypticC0der: im going back to try figure out the actual reason some processes find the env file empty, unsure as to why

crypticC0der:
tests passed but seemingly coverage tests broke?

mkalcok:
This approach seems like a bit of a band-aid solution, but I don't think that it's going to break anything and we do need the solution relatively fast, so I'm inclined to give it :+1:.
However, I think that we need to sort it out later in a more permanent way. The way I see it, we need to regenerate `ovn.env` file very time node leaves/joins the cluster, but right now we rely on manually calling `generateEnvironment` every time there's such operation. We need to tie these two operations together, which will require a bit of restructuring, because [`generateEnvironment`](https://github.com/canonical/microovn/blob/e5fb7def91e3ae81f0e10b489746542752f9dcc0/microovn/ovn/environment.go#L101) is not available in the same package as [`joinCentral`](https://github.com/canonical/microovn/blob/e5fb7def91e3ae81f0e10b489746542752f9dcc0/microovn/node/node.go#L221).
Additionally we need to ensure that when `northd` is started, `ovn.env` is up-to-date. This is also a functionality that could be baked into the [`joinCentral` function](https://github.com/canonical/microovn/blob/e5fb7def91e3ae81f0e10b489746542752f9dcc0/microovn/node/node.go#L221), what do you think?

> tests passed but seemingly coverage tests broke?

The coverage issue seems to be an infra/CI problem with APT repositories, not something that broke on our side. I expect that we can re-run the tests later and get positive results.

crypticC0der:
> Additionally we need to ensure that when northd is started, ovn.env is up-to-date. This is also a functionality that could be baked into the [joinCentral function](https://github.com/canonical/microovn/blob/e5fb7def91e3ae81f0e10b489746542752f9dcc0/microovn/node/node.go#L221), what do you think?
every time a node joins/leaves ovn.env is updated for any given node in that cluster

previously we did have a solution that only updates ovn.env when central is about to start, though we agreed that was somewhat of a hacky solution and moved to the current one, maybe this is a good discussion to have in person?

mkalcok:
As long as we agree that this needs a better long-term solution, I created [an LP issue](https://bugs.launchpad.net/microovn/+bug/2085546)

mkalcok:
I guess anything with `--wait=sb` will do (router, switch, doesn't really matter). But don't forget about cleanup. Either chose a unique name or make sure that the resources are removed on teardown.

mkalcok:
Btw, the issue with coverage was not a infra/CI issue, but a stale apt cache. #195 fixes it, and you can see that it tries to install newer version of go package, than the failed run in this PR.

crypticC0der:
sorry forgot to commit the tests

tests/test_helper/bats/cluster.bats
```diff
@@ -35,6 +35,9 @@ cluster_register_test_functions() {
     bats_test_function \
         --description "Open_vSwitch external_ids:ovn-remote addresses" \
         -- cluster_ovs_ovn_remote_addresses
+    bats_test_function \
+        --description "Ensure database connectivity" \
```
mkalcok, line 39: nit: The description is bit vague/off. This test is more about ensuring that northd processes data between NB and SB

tests/test_helper/bats/cluster.bats
```diff
@@ -208,4 +211,23 @@ cluster_test_db_connection_string() {
     done
 }
 
+# _test_southbound_propagation
```
mkalcok, line 214: nit: typo in the name

mkalcok (approved):
I have two nit picks, that I'll leave for your consideration. Otherwise LGTM, thanks.

tests/test_helper/bats/cluster.bats
mkalcok: The function is called `cluster_test_southbound_propagation`, you are missing the `cluster` part in the docstring. I see that similar "typo" is in the function above, but the other documented function in this file has the full name.
I'm not sure what's the reasoning behind leaving the "cluster" part out of the name, but I personally prefer having the full name in the docstring because it makes it easy to search for the function definition with `grep -iRn "# <function_name>" tests/`

crypticC0der: reasonable, saw it in the function above and assumed it was practice, should i change it in the function above, maybe in a seperate commit?

mkalcok (approved):
Thanks for the update. LGTM!

PR #198, opened 2024-11-01
snap: Skip golangci-lint on riscv64.

fnordahl:
The golangci-lint snap is unfortunately not built for riscv64.

The snapcraft advanced grammar [0] does unfortunately not allow us to express different build dependencies on different architectures without repetition of the list of dependencies.

Making use of composition at the YAML level is also not possible because merging sequences is just not supported [1].

0: https://snapcraft.io/docs/snapcraft-advanced-grammar
1: https://github.com/yaml/yaml/issues/35#issuecomment-496184265
Reported-at: alexmurray/golangci-lint-snap#3

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -245,8 +245,13 @@ parts:
       - ovn
       - ovs
     build-snaps:
-      - go/1.22/stable
-      - golangci-lint
+      # golangci-lint is currently not available on all architectures.
+      # (alexmurray/golangci-lint-snap#3)
+      - to riscv64:
+        - go/1.22/stable
+      - else:
+        - go/1.22/stable
+        - golangci-lint
     plugin: nil
     override-pull: |
       craftctl default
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -274,7 +279,11 @@ parts:
       export CGO_LDFLAGS_ALLOW="(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"
 
       # Check that `golangci-lint` is happy with the code.
-      golangci-lint run --verbose
+      if command -v golangci-lint; then
```
mkalcok, line 282: nit: We could use more explicit condition here based on [`CRAFT_ARCH_BUILD_ON `](https://snapcraft.io/docs/parts-environment-variables#heading--snapcraft-configuration-core24)

mkalcok (approved):
Overall LGTM, one optional suggestion inline

snap/snapcraft.yaml
fnordahl: That is indeed an option, I chose not to do that as we have to declare an assertion on architecture in the `build-snaps` already to avoid build failure on the back of the snap not being available:
```yaml
      - to riscv64:
        - go/1.22/stable
      - else:
        - go/1.22/stable
        - golangci-lint
```

With that in mind it felt a bit too much to repeat that assertion here.

PR #204, opened 2024-11-08
tests: Add optional parameters to 'microovn_add_vif'

fnordahl:
Two new optional parameters allow this function to create VIFs connected to specific LS and with specific IP address.


mkalcok (approved):
LGTM, thanks for bringing this to `main` and fixing `cidr` and `shift`

PR #205, opened 2024-11-12
Use RST for top-level docs, add contributing docs to the developer documentation.

fnordahl:


docs/developers/contributing.rst
```diff
@@ -10,25 +12,27 @@ use your contributions. In effect, you’re giving us a license, but you still
...
+* MicroOVN makes use of the GitHub Pull Request workflow.  There is no
   meaningful way to manage interdependencies between GitHub PRs, so we expect
   dependent changes proposed in a single PR reviewed and merged as separate
   commits.
-- A proposal for change is not complete unless it contains updates to
+* A proposal for change is not complete unless it contains updates to
   documentation and tests.
 
-## Tests
+Tests
```
mkalcok, line 34: Now that this page is part of the documentation, it would probably be better to link this section to `docs/developers/testing.rst`, instead of repeating the information.

mkalcok (changes requested):
I think that move of `CONTRIBUTING` into the docs makes sense. I have one suggestion in-line.

docs/developers/contributing.rst
fnordahl: That is a good idea, slight complication for users reading this from the repository, but I think the comment pointing at the location of the docs which is only visible for someone reading the source would solve that nicely.

docs/developers/contributing.rst
```diff
@@ -39,33 +43,19 @@ For impure functions, i.e. functions with side effects, if you find yourself
...
-    make check-lint
+Running tests
+~~~~~~~~~~~~~
 
-#### Functional tests
+..
+  The developer documentation is in the docs/developers directory.
 
-To run the entire suite in serial:
+Please refer to the document on :doc:`testing <testing>` to learn how to :ref:`run MicroOVN tests`.
```
mkalcok, line 58: Out of curiosity, what's the reason for these two links? They seem to point to pretty much the same location.

fnordahl: If you read this from `vim` you need to know which file to look in, and it might as well be a link for anyone reading from the web.

mkalcok (approved):
LGTM, thanks.

PR #126, opened 2024-03-07
Docs: how-to docs for building/installing/testing

mkalcok:
New how-to guides added for building, installing and testing of MicroOVN

Depending on: #124

PR #207, opened 2024-11-25
Remove 'microovn.' prefix in OVN commands

crypticC0der:
Automatic aliases have been enabled in the MicroOVN snap [1], so documentation has been updated to drop this prefix and rewrite certain small sections talking about how to run OVN commands through MicroOVN.

[1]: https://forum.snapcraft.io/t/alias-request-for-various-microovn-binaries/43755/5

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -92,6 +92,14 @@ linkcheck_anchors_ignore_for_url = [
 ]
 linkcheck_anchors_ignore_for_url.extend(custom_linkcheck_anchors_ignore_for_url)
 
+# this cisa.gov pages are so heavily refrenced the github CI for linkchecks is
+# being seen as a DDos, this is causing our linkcheck tests to fail so we have
+# to ignore it.
+linkcheck_ignore = [r'.*cisa\.gov.*']
+
+linkcheck_timeout = 120
+linkcheck_retries = 3
+
 # Tags cannot be added directly in custom_conf.py, so add them here
 for tag in custom_tags:
     tags.add(tag)
--- a/docs/reference/aliasing.rst
+++ b/docs/reference/aliasing.rst
@@ -1,0 +1,37 @@
+=================
+Automatic Aliases
+=================
+
+MicroOVN is distributed by snap, which has automatic aliases for OVN and OVS
+binaries. You can view these with the snap aliases command:
+
+.. code-block:: none
+
+    snap aliases microovn
+
+.. code-block:: none
+
+	Command                Alias         Notes
+	microovn.ovn-appctl    ovn-appctl    -
+	microovn.ovn-nbctl     ovn-nbctl     -
+	microovn.ovn-sbctl     ovn-sbctl     -
+	microovn.ovn-trace     ovn-trace     -
+	microovn.ovs-appctl    ovs-appctl    -
+	microovn.ovs-dpctl     ovs-dpctl     -
+	microovn.ovs-ofctl     ovs-ofctl     -
+	microovn.ovs-vsctl     ovs-vsctl     -
+	microovn.ovsdb-client  ovsdb-client  -
+	microovn.ovsdb-tool    ovsdb-tool    -
+
+Further inspection can be done by inspecting the files themselves:
+
+.. code-block:: none
+
+	ls $(which ovn-nbctl) -l
+
+.. code-block:: none
+
+	lrwxrwxrwx 1 root root 18 Nov 28 15:25 /snap/bin/ovn-nbctl -> microovn.ovn-nbctl
+
+These aliases are not related to the MicroOVN snap version and are managed by
+the store. All installations, done through the snap store, should have access to these aliases. This does mean if you install a locally built version of MicroOVN, these aliases are not created for you.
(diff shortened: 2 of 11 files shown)


mkalcok (changes requested):
Overall LGTM, thanks for the update. However I think that we would also benefit from either a reference or an explanation of the aliasing system, stating that OVN/OVS binaries can be executed either via `microovn.` prefixed commands or directly thanks to automatically created aliases (that can be inspected via `snap aliases`).
Also, since these auto aliases are not related to the microovn snap version, but rather they are something managed by the snap store, it would be good to find out if (when?) existing installations gain access to these aliases, and document that as well. Otherwise just dropping the prefixes in the documentation can lead to a confusion of users of existing installations.

fnordahl:
> Overall LGTM, thanks for the update. However I think that we would also benefit from either a reference or an explanation of the aliasing system, stating that OVN/OVS binaries can be executed either via `microovn.` prefixed commands or directly thanks to automatically created aliases (that can be inspected via `snap aliases`). Also, since these auto aliases are not related to the microovn snap version, but rather they are something managed by the snap store, it would be good to find out if (when?) existing installations gain access to these aliases, and document that as well. Otherwise just dropping the prefixes in the documentation can lead to a confusion of users of existing installations.

While I appreciate the quest for improving the documentation further, I wonder if we might be reaching into the responsibilities of generic snap documentation if we dive to deep into the alias topic. Would it perhaps be sufficient to mention that there are multiple ways to call binaries from the snap and link to https://snapcraft.io/docs/commands-and-aliases ?

On the topic of when the aliases come into effect, we have anecdotal evidence of them coming into effect on running systems without end user intervention, and from cursory view the `snapd` source code confirms this is the case:
https://github.com/canonical/snapd/blob/bd817d5d8c1d4732ad6a61346529dfe079ec963f/overlord/snapstate/snapstate.go#L2095

mkalcok:
> While I appreciate the quest for improving the documentation further, I wonder if we might be reaching into the responsibilities of generic snap documentation...

Perhaps a bit, but not too much imo. At least in the current state, auto-aliases are rather obscure. You won't find information about them neither on snapstore page, `snap info` nor in the code/`snapcraft.yaml`. So I think that the next best place is to document them in our docs.

> we have anecdotal evidence of them coming into effect on running systems without end user intervention

That is a good find, thank you. I think that this fact is well captured in this change with
```
These aliases are not related to the MicroOVN snap version and are managed by
the store. All installations should have access to these aliases.
```

docs/reference/aliasing.rst
```diff
@@ -0,0 +1,37 @@
+==================
+Automatic Aliasing
+==================
```
mkalcok, line 3: Small note, as per Peter's comments, we should avoid using gerunds in the page titles.
https://github.com/canonical/microovn/pull/126#issuecomment-1984172113

fnordahl:
> > While I appreciate the quest for improving the documentation further, I wonder if we might be reaching into the responsibilities of generic snap documentation...
> 
> Perhaps a bit, but not too much imo. At least in the current state, auto-aliases are rather obscure. You won't find information about them neither on snapstore page, `snap info` nor in the code/`snapcraft.yaml`. So I think that the next best place is to document them in our docs.
> 
> > we have anecdotal evidence of them coming into effect on running systems without end user intervention
> 
> That is a good find, thank you. I think that this fact is well captured in this change with
> 
> ```
> These aliases are not related to the MicroOVN snap version and are managed by
> the store. All installations should have access to these aliases.
> ```

The page @crypticC0der added about auto aliases looks fine to me. Let's keep it!

You have a fair point about the lack of documentation on the snapd / snap store side. I'd suggest we raise an issue over there about it.

@crypticC0der on the topic of the flaky CI, would you have a moment to experiment with [Sphinx linkcheck options](https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder) to see if we can get it to behave? I know that you have not changed links as part of this change, but issues like this often creep up on software repositories, and fixing creeping issues unrelated to your change is a great way to motivate reviewers to accept your change!

mkalcok:
I have a one small note that I realized today as I was working with locally built microovn. Auto-aliases work only when installed via snapstore (understandable). We could add a small note about this to `developers/building.rst`

docs/conf.py
```diff
@@ -91,6 +91,10 @@
     r'https://github\.com/.*'
 ]
 linkcheck_anchors_ignore_for_url.extend(custom_linkcheck_anchors_ignore_for_url)
+linkcheck_ignore = [r'.*cisa\.gov.*']
```
mkalcok, line 94: could we have a a short comment here explaining why we are skipping this check?

mkalcok (changes requested):
Thanks for the update. I have one nit inline. otherwise LGTM.

mkalcok (approved):
LGTM, thanks.

crypticC0der:
> The docs/conf.py file comes from [a template repository](https://github.com/canonical/sphinx-docs-starter-pack/blob/main/docs/conf.py). As a service to other travelers would you consider proposing a similar change there?

https://github.com/canonical/sphinx-docs-starter-pack/pull/317

PR #208, opened 2025-01-08
ci: Manually install LXD snap.

mkalcok:
It seems that the LXD snap is no longer installed by default on the ubuntu images in GH CI. Manual installation should fix that.

Note: This should be backported to all "live" branches.


crypticC0der (approved):
LGTM

PR #216, opened 2025-01-13
[branch-24.03] Makes tests run weekly

crypticC0der:
This makes the doc checks and tests run on mondays, the commit checks are not ran as they cannot begin failing when they once passed.


mkalcok (approved):
LGTM. thanks.

PR #218, opened 2025-01-23
microovn: Use constant for MicroCluster port.

fnordahl:
The default port for use by the MicroCluster is repeated three places in the microovn client command.

Let's replace those with a constant instead.

### DIFF
--- a/microovn/cmd/microovn/cluster_bootstrap.go
+++ b/microovn/cmd/microovn/cluster_bootstrap.go
@@ -43,7 +43,7 @@ func (c *cmdClusterBootstrap) Run(cmd *cobra.Command, args []string) error {
 
 	// Get system address.
 	address := util.NetworkInterfaceAddress()
-	address = util.CanonicalNetworkAddress(address, 6443)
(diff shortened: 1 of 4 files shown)


microovn/cmd/microovn/constants.go
mkalcok, line 1: nit: Creating new file for holding single constant seems superfluous. Especially if we don't foresee keeping more constants. Imo the port will have to become configurable at some point, so even this won't remain `const`. I'd suggest just moving it to the `main.go` for the time being.

mkalcok (dismissed):
Makes sense, overall LGTM. I left one nitpick inline for your consideration, but not a blocker.

microovn/cmd/microovn/constants.go
fnordahl: That does indeed make sense, my initial assumption was that this value would be used in multiple packages, but that turned out to not be the case, and I guess I stuck with a state of mind saying a separate file would be required. Thanks!

mkalcok (approved):
LGTM, thanks.

PR #219, opened 2025-01-23
Fixed bug "Commands with missing arguments exit with return code 0"

theovaka:
Fixed return value of commands when incorrect number of arguments is provided. Before the cmd.Help() function was called which returned 0 indicating success and now 1 is returned to indicate error.

Bug URL: https://bugs.launchpad.net/microovn/+bug/2076246


tests/test_help_functionality.bats
```diff
@@ -0,0 +1,94 @@
...
+        # Run the cluster disable command with invalid arguments
+        run lxc_exec "$container" "microovn disable"
+
+        # Assert the return code is 1
+        assert_failure
+
+    done
+}
+
+test_valid_args_return_0() {
```
mkalcok, line 80: I think this test may be bit superfluous. I don't have strong opinions on it, but I'd be OK with dropping it.

tests/test_help_functionality.bats
```diff
@@ -0,0 +1,94 @@
...
+        --description "Test invalid arguments return error code 1" \
+        -- test_invalid_args_return_1
+
+    bats_test_function \
+        --description "Test valid arguments return code 0" \
+        -- test_valid_args_return_0
+}
+
+test_invalid_args_return_1() {
+    for container in $TEST_CONTAINERS; do
```
mkalcok, line 34: I don't think it's necessary to run this on every cluster member. Since they are installed from same snap file, I think it's safe to assume that their behavior is the same.

tests/test_help_functionality.bats
```diff
@@ -0,0 +1,94 @@
...
+        # Run the cluster add command with invalid arguments
+        run lxc_exec "$container" "microovn cluster add"
```
mkalcok, line 36: nit: This is just for your consideration, not necessarily required. What would you think about defining the all the "bad commands" in an array and then just loop through them, instead of defining/running them individually. I think it would make it easier to extend this test in the future if new subcommands are added.

crypticC0der:
hm?

PR #224, opened 2025-01-28
Add commit message check

crypticC0der:
Adds a simple git workflow to check if the commit message and subject follow the 50/72 rule. This is part of the commits tests to be ran on PRs.

### DIFF
--- a/snapcraft/commands/switch.start
+++ b/snapcraft/commands/switch.start
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -eu
 
-. "${SNAP}/ovn.env"
+export OVS_RUNDIR="${SNAP_COMMON}/run/switch/"
 
 # Setup directories
 export OVS_LOGDIR="${SNAP_COMMON}/logs"
@@ -10,6 +10,8 @@ export OVS_DBDIR="${SNAP_COMMON}/data/switch/db"
 export OVS_PKGDATADIR="${SNAP}/share/openvswitch"
 export OVS_BINDIR="${SNAP}/bin"
 export OVS_SBINDIR="${SNAP}/bin"
+
+sudo mkdir -p ${SNAP_COMMON}run/switch ${SNAP_COMMON}/logs ${SNAP_COMMON}/data/switch/db
 
 # Start vswitchd
 "${SNAP}/share/openvswitch/scripts/ovs-ctl" \
--- a/snapcraft/commands/switch.stop
+++ b/snapcraft/commands/switch.stop
@@ -1,8 +1,7 @@
 #!/bin/sh
 set -eu
 
-. "${SNAP}/ovn.env"
-
+export OVS_RUNDIR="${SNAP_COMMON}/run/switch/"
 # Stop vswitchd
 "${SNAP}/commands/ovs-appctl" exit --cleanup
 "${SNAP}/share/openvswitch/scripts/ovs-ctl" stop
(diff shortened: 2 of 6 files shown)


.github/workflows/commits.yml
```diff
@@ -46,3 +46,34 @@ jobs:
...
+
+            # Check subject line length
+            if [ ${#subject} -gt 50 ]; then
+              echo "Commit subject exceeds 50 characters: '$subject'"
+              exit 1
+            fi
+
+            # Check body line length
+            echo "$body" | while read -r line; do
+              if [ ${#line} -gt 72 ]; then
```
mkalcok, line 74: Sometimes a line above 72 chars is necessary, for example if you want to add link as a reference ([example](https://github.com/canonical/microovn/commit/ba489f9cf61594649c4ca8b5906e74aeecdb7cdd)). I wonder if there are off the shelf solutions for commit message checks that cover these corner cases :D

mkalcok (changes requested):
Nice QOL idea. Overall I'm +1 on it. I left one comment inline, plus some general thoughts here:

If we are going to start enforcing this, we should state these requirements in the contribution docs.

Not a blocker, just something to think about. I think it would be better to have this logic in the main codebase (bash script/make target?). GH action could just run the script. That way it would be runnable locally and it would also simplify migration if we ever need to move to different CI.

crypticC0der:
@mkalcok could a final review be done

tools/commitcheck.sh
```diff
@@ -0,0 +1,21 @@
...
+		# Check subject line length
+		if [ ${#subject} -gt 50 ]; then
+			echo "Commit subject exceeds 50 characters: '$subject'"
+			exit 1
+		fi
+		# Check body line length
+		echo "$body" | while read -r line; do
+			if [ ${#line} -gt 72 ]; then 
+				# checks if the line is of the form <1-20 chars>: link
+				if ! echo "$line" | grep -Eq '.{1,20}:\ (http(s)?://.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}([-a-zA-Z0-9@:%_\+.~#?&//=]*)' ; then
```
mkalcok, line 14: Few notes to this regexp:

I can't find any authoritative sources, but isn't the usual way to annotate links
```
[0] <link>
```
What would be the `<1-20 chars>:` in your suggestion?

Is there an added value in having this complicated regexp trying to match intricacies of a valid URL? Wouldn't something easier be sufficient? For example (just a suggestion) `^\[[0-9]+\] http(s)*://`

mkalcok (changes requested):
I mentioned couple nits inline, but main blocker right now is lack of documentation. As mentioned in previous review, if we are going to enforce this commit formatting, then it should be mentioned at least in the contributing guide.

tools/commitcheck.sh
crypticC0der: for some commits we have Reported-at: <link>
and so i thought anything under 20 characters covers any situation in which we need to reference a link.
also regarding the url regex, I just used the one I have on my regex cheatsheet, i dont mind changing it to simpler query

mkalcok: > for some commits we have Reported-at:

I see, now that you mention it, we should probably disregard all tags for this check. For example it's also likely that `Signed-off: NAME <EMAIL>` can exceed our lien limit.

crypticC0der:
Unsure about the change to the contributing page, maybe it should also mention something about how to add links, or maybe it should explain the 50/72 rule?

PR #231, opened 2025-04-11
microovn paths: single source of truth

crypticC0der:
moves microovn towards a single source of truth which was the original goal for microovn paths

### DIFF
--- a/microovn/cmd/microovn/paths.go
+++ b/microovn/cmd/microovn/paths.go
@@ -16,16 +16,23 @@ func (c *cmdPath) Command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "path <NAME>",
 		Short: "Retrieve the path for the given name.",
-		Args:  cobra.ExactArgs(1),
+		Args:  cobra.RangeArgs(0, 1),
 		RunE:  c.Run,
 	}
 	return cmd
 }
 
 func (c *cmdPath) Run(_ *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		for k, v := range paths.PathsMap {
+			fmt.Printf("%s: %s\n", k, v)
+		}
+		return nil
+	}
+
 	name := args[0]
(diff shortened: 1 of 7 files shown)


crypticC0der:
as a note I do not think a single source of truth is worth the annoyance (however minor) of microovn paths, currently in the codebase there is minimal code reuse, which from what ive seen writing this patch could be minimized even more (by things such as chassis.start reading from ovn-central.env)
Paths has some annoyances and would need to be expanded to include every possible path if thats the direction we go down as it presently only uses the paths defined previously in microovn/ovn/paths.

I think reverting #229 and not merging this PR would be a preferable decision to embracing microovn path, not because the code itself is bad, but just because it over-complicates things to solve a minor case of multiple sources of truth

fnordahl:
Thanks a lot for executing the experiment @crypticC0der, which sprung out from this review comment from @mkalcok: https://github.com/canonical/microovn/pull/228#discussion_r2013951306

> I think reverting #229 and not merging this PR would be a preferable decision to embracing microovn path, not because the code itself is bad, but just because having it over-complicates things to solve a minor case of multiple sources of truth

I tend to agree with you, while I still agree with the original intent of avoiding repeating the same path string in multiple places (which in isolation is tech debt waiting to grow into a bug), this approach has proven to not be as straight forward.

These paths and the names used to identify them are really internal data. For a user visible CLI command, there would be additional unanswered questions to really complete it, such as docs. And how would we explain these internal names to a user without just creating more confusion?


As an alternative we could move all of this to the build system, renaming scripts to `*.in` and do a template based approach to them. Similarly we can pass these values into the Go code through [build time flags](https://github.com/canonical/microovn/blob/336b7904ac053fd9a8eed2936f241e217e0510b3/snap/snapcraft.yaml#L316-L318).

For now, let's file a bug to capture this and move on.

fnordahl:
As discussed, closing this, remember to raise a revert of #229 @crypticC0der

PR #232, opened 2025-04-17
docs: add annotations to microcloud integration only

minaelee:
The MicroOVN is pulled in as part of the [MicroCloud integrated docs set](https://canonical-microcloud.readthedocs-hosted.com/en/latest/microceph/), which allows the user to switch between the MicroCloud, LXD, MicroCeph, and MicroOVN products from a header menu.

For MicroCloud users, some pages of the other docs sets are not relevant (such as installation docs, which MicroCloud handles). This PR adds a `note`-type annotation to such pages advising users of this, which uses the `only` Sphinx directive to _only_ show those notes when the page is viewed as part of the MicroCloud integrated docs set. It has no effect on the standalone doc set.

### DIFF
--- a/docs/tutorial/multi-node.rst
+++ b/docs/tutorial/multi-node.rst
@@ -2,6 +2,12 @@
 Multi-node
 ===========
 
+.. only:: integrated
+
+   .. note::
+
+      MicroCloud users can disregard the instructions on this page, because the MicroCloud setup process handles MicroOVN installation.
+
 This tutorial shows how to install a 3-node MicroOVN cluster.
 
 One big advantage of a multi-node cluster is that it provides redundancy
--- a/docs/tutorial/single-node.rst
+++ b/docs/tutorial/single-node.rst
@@ -2,6 +2,12 @@
 Single-node
 ===========
 
+.. only:: integrated
+
+   .. note::
+
+      MicroCloud users can disregard the instructions on this page, because the MicroCloud setup process handles MicroOVN installation.
+
 This tutorial shows how to install MicroOVN in the simplest way possible.
 
 .. caution::


mkalcok (changes requested):
Hi @minaelee, thanks for the documentation update. I assume that this is already established approach for all microcloud projects? I have few suggestions below on establishing this practice within microovn project. Let me know what you think.

On its own, the `.. only:: integrated` is a bit cryptic for someone without explicitly knowing that it's used in downstream documentation. What do you think about adding a page in the [developer's documentation section](https://github.com/canonical/microovn/tree/main/docs/developers) that would explain what the `integrated` tag is and how/when to use it.

If the MicroOVN contributors are expected to proactively use this tag when contributing documentation, It should be probably mentioned in the [contributing.rst](https://github.com/canonical/microovn/blob/main/docs/developers/contributing.rst). The above mentioned doc page should then also contain clear criteria on when to mark page/section as "not interesting for microcloud users".

I also briefly ran through the documentation and here are some pages/sections that are imo also candidates for this tag. These are just suggestions, ultimately, I'll leave it up to you:
* [Upgrade DB cluster from plaintext to TLS](https://github.com/canonical/microovn/blob/336b7904ac053fd9a8eed2936f241e217e0510b3/docs/how-to/tls.rst?plain=1#L152) - This concerns only very old installations of microovn and I don't think that microcloud ever used such old versions.
* [Setting up custom underlay network](https://github.com/canonical/microovn/blob/main/docs/how-to/ovn-underlay.rst) - This is only relevant when cluster is bootstrapped manually. Microcloud does this for the user during setup.

nit: I'd be OK with this change as a single commit. I don't think that splitting each file change into individual commit is necessary.

PR #233, opened 2025-04-30
Experimental/25.03 bgp

crypticC0der:


fnordahl (approved):
LGTM, thanks!

If i were to be pedantic I'd request cleaning the commit history a bit as it contains one WIP and one effectively becoming noop as patch changes are later dropped.

However this is a feature branch and we've already acknowledged that rework is required in multiple areas before integration to main, so three will be a manual curated merge job at the end.

Thank you for getting us unlocked and for progressing this!

PR #234, opened 2025-05-01
Makes IPv4 optional for BGP external connections.

crypticC0der:
Allows easier enabling of BGP via no longer needing IPv4. This is part of the attempt to get BGP unnumbered working within OVN. This configures the router-id which is generated based off a hash of the logical router port name.

### DIFF
--- a/microovn/frr/bgp/persistance.go
+++ b/microovn/frr/bgp/persistance.go
@@ -35,13 +35,7 @@ func EnsureInterfacesInVrf(ctx context.Context, s state.State) error {
 		}
 		portVrf = strings.TrimSpace(portVrf)
 
-		portIP, err := ovnCmd.VSCtl(ctx, s, "get", "port", port, fmt.Sprintf("external-ids:%s", BgpIfaceIP))
-		if err != nil {
-			allErrors = errors.Join(allErrors, fmt.Errorf("failed to lookup IPv4 address of port '%s': %v", port, err))
-		}
-		portIP = strings.Trim(strings.TrimSpace(portIP), "\"")
-
-		err = moveInterfaceToVrf(ctx, port, portIP, portVrf)
+		err = moveInterfaceToVrf(ctx, port, portVrf)
 		if err != nil {
 			allErrors = errors.Join(allErrors, fmt.Errorf("failed to move interface '%s' to VRF '%s': %v", port, portVrf, err))
 		}
--- a/microovn/frr/bgp/redirect.go
+++ b/microovn/frr/bgp/redirect.go
@@ -30,12 +30,6 @@ const BgpManagedTag = "microovn-bgp-managed"
 // ports be assigned
 const BgpVrfTable = "microovn-bgp-vrf"
 
-// BgpIfaceIP - a key used in "external_ids" table of OVS ports used for
-// BGP redirecting. It keeps track of the IPv4 address that should be
-// assigned to the port, to match IPv4 address of the Logical Router Port
-// from which the traffic is redirected
-const BgpIfaceIP = "microovn-bgp-ip"
-
 // BgpBridgeMapping - a key used in "external_ids" of Open_vSwitch table, to
 // keep track of "ovn-bridge-mappings" managed by MicroOVN.
 const BgpBridgeMapping = "microovn-bgp-bridge-mapping"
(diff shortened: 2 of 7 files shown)


mkalcok (changes requested):
Thanks for porting this change to 25.03.
I'm not sure I agree with taking away the ability to enable BGP over IPv4. In my opinion, the IPv4 parameter should be made optional, rather than completely taken away. For one, the functionality is already in there, and we know that there are deployments out there that utilize things like IPv4 link local addresses to setup BGP peer relations.

microovn/frr/bgp/redirect.go
```diff
@@ -112,6 +112,19 @@ func generateLrpMac(ifaceName string) string {
 	return strings.TrimRight(macAddr, ":")
 }
 
+// generateRouterID returns a router-id address based on a string. The returned
+// router-id will always be same for given interface name.
+// Warning: There is no guarantee that the address won't conflict with other
+// router-ids present in the AS.
+func generateRouterID(s string) string {
```
mkalcok, line 119: Function name and/or the documentation could clarify that this is pertains to BGP router ID. Not to be confused with, for example, OVN's Logical Router's ID.

microovn/frr/bgp/persistance.go
```diff
@@ -35,11 +35,12 @@ func EnsureInterfacesInVrf(ctx context.Context, s state.State) error {
...
-		if err != nil {
-			allErrors = errors.Join(allErrors, fmt.Errorf("failed to lookup IPv4 address of port '%s': %v", port, err))
-		}
-		portIP = strings.Trim(strings.TrimSpace(portIP), "\"")
+		portIP := ""
+		// portIP, err = ovnCmd.VSCtl(ctx, s, "get", "port", port, fmt.Sprintf("external-ids:%s", BgpIfaceIP))
+		// if err != nil {
+		// 	allErrors = errors.Join(allErrors, fmt.Errorf("failed to lookup IPv4 address of port '%s': %v", port, err))
+		// }
+		// portIP = strings.Trim(strings.TrimSpace(portIP), "\"")
```
mkalcok, line 43: This looks like it always disregards IP address of the port, even if it was previously set.

tests/test_helper/bats/bgp_data_plane.bats
```diff
@@ -32,10 +32,9 @@ function ping_ovn_int_network_over_bgp_router() {
 
     # Enable BGP redirection and start BGP daemon in OVN chassis
     local host_asn=4210000000
-    local BGP_NET_IP="172.16.10.1/24"
     local vrf="10"
     local vrf_device="ovnvrf$vrf"
-    local external_connections="$OVN_CONTAINER_INT_IFACE:$BGP_NET_IP"
+    local external_connections="$OVN_CONTAINER_INT_IFACE"
```
mkalcok, line 37: I think we should test both scenarios. i.e. with and without the explicit IPv4 address for the BGP interface.

mkalcok (changes requested):
Thanks for the update. I left few minor comments below, otherwise LGTM.

tests/test_helper/bats/bgp_data_plane.bats
crypticC0der: the ping_ovn_int_network_over_bgp_router function, doesnt undo anything it does so having a similar function that runs after it would be difficult, at present the control plane test sets up a bgp environment with ipv4 so the ipv4 bgp functionality is still being tested.

microovn/frr/bgp/persistance.go
```diff
@@ -35,8 +35,9 @@ func EnsureInterfacesInVrf(ctx context.Context, s state.State) error {
 		}
 		portVrf = strings.TrimSpace(portVrf)
 
-		portIP, err := ovnCmd.VSCtl(ctx, s, "get", "port", port, fmt.Sprintf("external-ids:%s", BgpIfaceIP))
-		if err != nil {
+		portIP := ""
+		portIP, err = ovnCmd.VSCtl(ctx, s, "get", "port", port, fmt.Sprintf("external-ids:%s", BgpIfaceIP))
+		if err == nil {
```
mkalcok, line 40: Is this condition correct? Wouldn't that mean that when `VSCtl` successfully finds interface's IP in the `external-ids`, we'd log an error?

tests/test_helper/bats/bgp_data_plane.bats
mkalcok: The required teardown changes are not that bad. Here's a crude attempt that mostly repurposes teardown function from control_plane tests: https://gist.github.com/mkalcok/4ace5c926404dff892b64f86f40209d8

mkalcok:
Apologies for my first review. I now realized/remembered that IPv4 parameter for the `external_connections` wasn't functional. It was just a crutch because OVN LRPs required an IP address. We never really fully tested/supported setting up BGP with explicit peers, we always tested only the "unnumbered" mode (IPv4 over IPv6 nexthop).

So with that said, I think we can go the route of removing the IPv4 from `external_connections` completely.

mkalcok:
The controlplane error seems to be caused by the `microovn_start_bgp_unnumbered` which manually configures FRR BGP daemon if `asn` argument is not supplied to `microovn enable bgp`.
That function does not explicitly set `router-id` and so far it relied on the IPv4 address of the interface.

Snippet of the FRR BGP configuration from test logs
```
     Local Role: undefined
     Remote Role: undefined
   Hostname: microovn-bgp-peer-1-1
     BGP version 4, remote router ID 0.0.0.0, local router ID 0.0.0.0
     BGP state = Idle
```

Updating the function explicitly set `router-id` fixes the issue.

crypticC0der:
should i make a quickfix forcing 2.2.0 snowballstreamer or do we wait for it to calm down?

mkalcok:
Pinning `snowballstemmer<3` (with reference to the bug report) might be our best option now.

mkalcok:
snowballstemmer 3.0 was yanked from pypi. I'm retriggering docs tests and then this PR should be good to go.

mkalcok (approved):
LGTM.

PR #235, opened 2025-05-08
Pin snowballstemmer to 2.2.0

crypticC0der:
hotfix for https://github.com/sphinx-doc/sphinx/issues/13533

### DIFF
--- a/docs/.sphinx/requirements.txt
+++ b/docs/.sphinx/requirements.txt
@@ -12,3 +12,4 @@ sphinx-reredirects
 sphinx-tabs
 sphinxcontrib-jquery
 sphinxext-opengraph
+snowballstemmer<3


mkalcok:
snowballstemmer 3.0.0 is [getting yanked from the pypi](https://github.com/snowballstem/snowball/issues/229#issuecomment-2862760480). This might not be needed.

crypticC0der:
ah thats good to know

flying-sheep:
As Henry said:

> Reminder to everyone, pinning is not needed now that it's yanked. **Also, you don't need to pin, you could have just used `snowballstemmer!=3.0.0`**, which protects against the broken version, but won't come back to bite you later when something starts requiring >=3.

mkalcok:
That's a good tip @flying-sheep, thanks.

PR #237, opened 2025-05-08
WIP: Learn default gateway from the network.

fnordahl:
Learning was implemented as part of the route-exchange work that went into OVN 25.03.  Let's make use of it in our functional tests!


mkalcok (approved):
It's great to see that the route learning just works (TM).

PR #238, opened 2025-05-09
Activates needed services on daemon start

crypticC0der:
when microovn.daemon is started it will be stuck in a error state if any of its needed services are not active, this is most easily noticeable where if you disable microovn and then re-enable it. This is solved by iterating through the expected microovn services and enabling their corresponding snap services.

fixes: https://launchpad.net/bugs/2101937

### DIFF
--- a/tests/test_helper/bats/cli_ovsovn.bats
+++ b/tests/test_helper/bats/cli_ovsovn.bats
@@ -85,6 +85,9 @@ cli_ovsovn_register_test_functions() {
     bats_test_function \
         --description "Test paths command"\
         -- test_paths
+    bats_test_function \
+        --description "Test enable disable microovn"\
+        -- test_disable_microovn
 }
 
 ovs-appctl_ovs-vswitchd() {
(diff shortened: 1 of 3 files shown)


tests/test_helper/bats/cli_ovsovn.bats
```diff
@@ -284,4 +287,22 @@ test_paths(){
     done;
 }
 
+test_disable_microovn(){
+    for container in $TEST_CONTAINERS; do
+        run lxc_exec $container "snap disable microovn"
+        assert_success
+        run lxc_exec $container "snap enable microovn"
+        assert_success
+        sleep 5
```
fnordahl, line 296: We should avoid sleeps in tests whenever possible, could this perhaps be replaced by the `microovn waitready` command?

fnordahl (approved):
Thanks for your work on this, @crypticC0der !

While actively managing the underlying systemd services is a big hammer to swing, I think it makes sense given the UX expected from making use of `snap disable microovn` and `snap enable microovn`.

I have a nit in-line, but I'll merge regardless to close this out, please post a follow-on when convenient.

PR #239, opened 2025-05-13
Allow remote enabling and disabling of services

crypticC0der:
Gives an optional node argument which find the related node if it exists, and then disables/enables the related service.


fnordahl (changes requested):
The new feature is in need of documentation.

microovn/api/services/control.go
```diff
@@ -72,6 +72,18 @@ func disableService(s state.State, r *http.Request) response.Response {
 		logger.Errorf("Failed to get service: %s", err)
 		return response.ErrorResponse(500, "internal server error")
 	}
+
```
mkalcok, line 75: Is this change supposed to be committed? It doesn't seem to have any effect aside from the log messasges.

microovn/client/client.go
```diff
@@ -264,3 +265,25 @@ func DeleteConfig(ctx context.Context, c *client.Client, key string) (types.Dele
...
+			cli, err := m.RemoteClient(clusterMember.Address.String())
+			if err != nil {
+				return nil, err
+			}
+
+			return cli, nil
+		}
+	}
+
+	return nil, errors.New("no such node")
```
mkalcok, line 288: This error message could use more context. It could contain the name of the node that wasn't found.

mkalcok (approved):
I could've swear that I reviewed this change some time ago, but I don't see my comments anywhere. Thank you for this improvement to users' QOL. I like the docs and test coverage, but I do have one note for the implementation.

MicroCluster provides native way of forwarding requests to to other nodes in the cluster if the endpoint has [`ProxyTarget` set to `true`](https://pkg.go.dev/github.com/canonical/microcluster/v2@v2.2.0/rest#EndpointAction), and we do have it on the service control endpoints:
https://github.com/canonical/microovn/blob/827b7487924a72850c697fb5dc1222ba12d111c5/microovn/api/services/control.go#L23-L24

This means that we could drop the manual client search in the `GetClientFromName` function and instead leverage the existing [`Target` method on the api URL](https://pkg.go.dev/github.com/canonical/lxd/shared/api#URL.Target).
Example:
```diff
diff --git a/microovn/client/client.go b/microovn/client/client.go
index 7996580..70dfbd0 100644
--- a/microovn/client/client.go
+++ b/microovn/client/client.go
@@ -194,11 +194,11 @@ func DisableService(ctx context.Context, c *client.Client, serviceName string, a
 
 // EnableService sends request to disable service with name as as specified in
 // "serviceName" argument.
-func EnableService(ctx context.Context, c *client.Client, serviceName string, extraConfig *types.ExtraServiceConfig) (types.WarningSet, types.RegenerateEnvResponse, error) {
+func EnableService(ctx context.Context, c *client.Client, serviceName string, extraConfig *types.ExtraServiceConfig, target string) (types.WarningSet, types.RegenerateEnvResponse, error) {
        queryCtx, cancel := context.WithTimeout(ctx, time.Second*30)
        defer cancel()
        scr := types.ServiceControlResponse{}
-       err := c.Query(queryCtx, "PUT", types.APIVersion, api.NewURL().Path("service", serviceName), extraConfig, &scr)
+       err := c.Query(queryCtx, "PUT", types.APIVersion, api.NewURL().Path("service", serviceName).Target(target), extraConfig, &scr)
        if err != nil {
                return types.WarningSet{}, types.RegenerateEnvResponse{}, fmt.Errorf("failed to enable service '%s': '%s'", serviceName, err)
        }
```

What do you think?

mkalcok (approved):
LGTM. thanks for the update.

fnordahl (approved):
LGTM, thanks!

PR #242, opened 2025-06-20
Allows set-ca to use stdin

crypticC0der:
This allows certificates set-ca to take certificates and keys from stdin, this is very useful for ongoing charm work and general consumption of this feature.

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -69,3 +69,4 @@ unencrypted
 uptime
 vSwitch
 yaml
+stdin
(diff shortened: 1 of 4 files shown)


fnordahl (changes requested):
Allowing cert or key or both to be provided through STDIN using the existing command line arguments increases complexity of the code, and there appears to be a couple of unchecked error conditions that need to be addressed if you insist on this approach.

I wonder if it would make it simpler if you leave the existing arguments as is and add a third `--combined` argument (or similar?) which only supports reading from STDIN and expects all data to be present?

Then the argument parsing code can assert that either, both `--cert` and `--key`, or `--combined` alone is provided.

In any cased an update to `User-provided CA certificates` section of docs/how-to/tls.rst is required.

fnordahl (approved):
LGTM, thanks!

PR #246, opened 2025-07-09
CI: Limit TICS job execution.

mkalcok:
TICS job requres authentication token. Don't attempt to run
TICS-related CI jobs if the token is not accessible.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -248,6 +248,12 @@ jobs:
       SRC_DIR: ${{ github.workspace }}/microovn/
     needs:
       - generate-coverage
+    # TICS execution requires access to an auth. token which is not available in the forks of this
+    # repository. We allow this job to run only if it has access to the token:
+    #  * On "push" action
+    #  * On scheduled action
+    #  * On pull request created from the main repository (not fork)
+    if: ${{ (github.event_name != 'pull_request') || (github.repository == github.event.pull_request.head.repo.full_name) }}
 
     steps:
       - name: Checkout code


.github/workflows/tests.yml
```diff
@@ -298,17 +298,3 @@ jobs:
           viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=GoProjects
           ticsAuthToken: ${{ secrets.TICSAUTHTOKEN }}
           installTics: true
-
-      # This type of the TICS run compares current state of the code against the last
-      # persistent measurement point. We run in on proposed pull requests.
-      - name: "Comparative TICS run (mode: client)"
-        if: ${{ github.event_name == 'pull_request' }}
```
fnordahl, line 305: Could we make it run when there is access to the token instead of removing it?

As @azkaratea pointed out OOB, the MIR project appears to have an example of this:
https://github.com/canonical/mir/blob/863945876865f272ae6be3c0ccf4eaaa0d64ea29/.github/workflows/tics.yml#L20-L21

fnordahl (changes requested):
> PRs from forked repositories don't have access to the
secret token required to run TICS. As a result, the step
that triggered TICS run in "client" mode on PRs is useless.

While there is an opportunity to improve this so that third party PRs can access this feature, I don't agree that this is useless.

People with access to the canonical organization can still make use of it by creating PRs from branches on the microovn repository, so I'd rather keep it.

mkalcok:
Thanks for the review @fnordahl And you are right, description and title needed update.

fnordahl (approved):
LGTM, thanks!

PR #247, opened 2025-07-10
Datapath mode

mkalcok:
Work in progress to allow MicroOVN to operate in datapath-only mode with external ovn-central services.

please review commit-by-commit

### DIFF
--- a/docs/how-to/datapath-only-mode.rst
+++ b/docs/how-to/datapath-only-mode.rst
@@ -1,0 +1,135 @@
+==================
+Datapath-only Mode
+==================
+
+This is a mode of operation in which MicroOVN cluster runs only ``switch`` and
+``chassis`` service, with ``central`` service being provided externally by another
+MicroOVN cluster or by some other OVN deployment.
+
+.. important::
+
+   This is not an equivalent of the OVN Interconnect feature. It does not allow
+   connecting two full OVN deployments. Instead it is meant to help with migrations,
+   or in hybrid deployments, where OVN central services are deployed by some other
+   method.
+
+Cluster setup
+-------------
+
+We will start by deploying the MicroOVN cluster without the ``central`` service. Our
+hosts that run MicroOVN are called ``node-1``, ``node-2`` and ``node-3``. Setup of
+the external OVN central cluster is not covered by this how-to. For the sake of this
+guide we assume that there are OVN Southbound and OVN Northbound services running on
+their default ports on IPs:
+
+* 10.0.0.1
+* 10.0.0.2
+* 10.0.0.3
+
+.. note::
+
+   If you already have MicroOVN deployed, see
+   :doc:`Service Control </how-to/service-control>` about how to disable ``central``
+   service on the running nodes, and :doc:`MicroOVN services </reference/services>`
+   regarding implications of removing all ``central`` services on the running cluster.
+
+Obtain CA certificate and private key for the MicroOVN
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Because the MicroOVN enforces encrypted communication between OVN services, we need
+to ensure that OVN services running on the MicroOVN cluster and on the external OVN
+central cluster are mutually trusted.
+
+Obtain intermediate CA certificate and private key from the CA that issued certificates
+for the external OVN central cluster and place it in the ``/var/snap/microovn/common/``
+on the ``node-1``, under names ``ca.cert`` and ``ca.key``.
+
+Bootstrap MicroOVN cluster
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First we bootstrap the MicroOVN cluster on the ``node-1`` and generate tokens, so that
+``node-2`` and ``node-3`` can join.
+
+.. code-block::
+
+   sudo microovn init
+
+We will be taken through an initialisation process that will look something like this:
+
+.. code-block::
+
+   Please choose the address MicroOVN will be listening on [default=10.75.224.213]:
+   Would you like to create a new MicroOVN cluster? (yes/no) [default=no]: yes
+   Please select comma-separated list services you would like to enable on this node (central/chassis/switch) or let MicroOVN automatically decide (auto) [default=auto]: switch,chassis
+   Please choose a name for this system [default=node-1]:
+   Would you like to define a custom encapsulation IP address for this member? (yes/no) [default=no]:
+   Would you like to provide your own CA certificate and private key for issuing OVN TLS certificates? (yes/no) [default=no]: yes
+   Please enter the path to the CA certificate file: /var/snap/microovn/common/ca.cert
+   Please enter the path to the CA private key file: /var/snap/microovn/common/ca.key
+   Would you like to add additional servers to the cluster? (yes/no) [default=no]: yes
+   What's the name of the new MicroOVN server? (empty to exit): node-2
+   <NODE_2_TOKEN>
+   What's the name of the new MicroOVN server? (empty to exit): node-3
+   <NODE_3_TOKEN>
+
+Then we join the cluster with ``node-2`` and ``node-3``
+
+.. code-block::
+
+   sudo microovn init
+
+Similar dialogue takes us through the joining process, the most notable difference is that
+we will select ``no`` when asked about whether we want to create a new cluster.
+
+.. code-block::
+
+   Please choose the address MicroOVN will be listening on [default=10.75.224.175]:
+   Would you like to create a new MicroOVN cluster? (yes/no) [default=no]: no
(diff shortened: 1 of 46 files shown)


microovn/ovn/environment/environment.go
```diff
@@ -220,3 +221,42 @@ func CleanupPaths() error {
 
 	return errors.Join(errs...)
 }
+
+// UpdateOvnListenConfig configures the OVN NB and SB databases to listen on the appropriate ports.
+func UpdateOvnListenConfig(ctx context.Context, s state.State) error {
```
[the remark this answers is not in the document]
mkalcok: To me it seemed appropriate to have these function in a package that manipulates environment to alter OVN's runtime behavior, but I also see your point that this file was previously dedicated more to the environment variables and cli argument values.

On the other hand, i kinda like that the "cmd" package exposes only generic wrappers for running general OVN cli commands.

Perhaps a new package would be a solution here?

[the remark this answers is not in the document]
fnordahl: Yes new package also works for me. Composition FTW!

microovn/cmd/microovn/init.go
```diff
@@ -115,6 +115,16 @@ func (c *cmdInit) Run(_ *cobra.Command, _ []string) error {
 		}
 
 		optionalConfig := make(map[string]string)
+
+		// TODO: Perhaps we need to change this question to get comma separated values
+		selectedServices, err := c.common.asker.AskChoice("Please select services you would like to enable on this node (central/chassis/auto) [default=auto]: ", []string{"central", "chassis", "auto"}, "auto")
```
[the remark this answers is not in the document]
mkalcok: yeah, I was thinking about this and I was wondering if there's a practical use-case for running switch service only. Should we perhaps consolidate roles to only `central` and `chassis`?

Note: There are pieces of code dating back to original implementation that treat switch/chassis service role interchangeably, those should be probably fixed if we keep it. (It shouldn't be too much work)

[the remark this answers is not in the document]
fnordahl: There is a practical use case for running switch service only.

The MicroOVN snap is our preferred distribution channel for Open vSwitch and OVN, when in use it is the sole source of Open vSwitch on the sytem.

When using the userspace data path (DPDK, OVS-DOCA), or kernel mode hardware offlload a Open vSwitch bridge will be responsible for all networking for a host with no oob management. And yes there are customers with these requirements.

https://bugs.launchpad.net/microovn/+bug/2045490
https://github.com/canonical/netplan/pull/549
https://docs.openstack.org/project-deploy-guide/charm-deployment-guide/latest/install-maas.html#create-ovs-bridge

fnordahl (changes requested):
Thank you for your swift work on this!

This is not a full review, although I've read through the changes to the processing of internal vs. external source of SB DB connection string, and the approach seems reasonable to me.

I have an ask for namespacing the config command, but I only expect us to do this in the CLI level for now, see in-line comment.

microovn/node/node.go
```diff
@@ -40,13 +41,16 @@ func DisableService(ctx context.Context, s state.State, service types.SrvName) e
...
 	// and if this check is moved into the later "if central" then we
 	// will need to handle the database check on each branch of the if
+	lastCentral := false
 	if service == types.SrvCentral {
 		centrals, err := FindService(ctx, s, service)
 		if err != nil {
 			return err
 		}
+		// TODO: We should probably have some --i-really-mean-it flag
+		// from the client request before allowing this
```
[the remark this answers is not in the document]
mkalcok: anti-foot-gun measures were applied by requiring `--allow-disable-last-central` flag to disable last central node.

microovn/ovn/environment/environment.go
```diff
@@ -17,7 +17,6 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microcluster/v2/cluster"
 	"github.com/canonical/microcluster/v2/state"
-
```
fnordahl, line 20: Is removal of this blank line intentional?

It provided logical separation of imports, which I think we'd like to keep.

microovn/ovn/join.go
```diff
@@ -48,9 +48,15 @@ func Join(ctx context.Context, s state.State, initConfig map[string]string) erro
...
+	// central only if:
+	//   * external OVN central wasn't configured
+	//   * or if there are less than 3 MicroOVN nodes with 'central' service enabled
+	externalOvnCentral, err := environment.IsExternalCentralConfigured(ctx, s)
+	if err != nil {
+		return err
+	}
 	enableServices := requestedServices{
-		Central: srvCentral < 3,
+		Central: !externalOvnCentral && srvCentral < 3,
```
fnordahl, line 59: Thank you for adding this, it's a nice compromise to expedite charm development in canonical/microovn-operator#21

fnordahl (commented):
Thank you for your swift and thorough work on this, it looks good to me, pending some updates to commit messages, newline changes and docs.

I love the test coverage!

I also like the approach of configuring the IPs separately, which both makes the interface simpler for a human to understand (they do not need to know the intricate details of OVSDB connection strings to use it), and secure by default by enforcing ssl.

There are a few commits which need updating of subject (`WIP` prefix) and content (naming changed during review)to match the actual state of implementation, but I assume you will update these when you add docs.

microovn/ovn/environment/environment.go
mkalcok: Thanks for catching this. It wasn't intentional.

mkalcok:
Thank you for the review.
Yes the `WIP` tags in commits mark those that still need doc update, just so I won't forget anything :laughing:

I'll double-check for naming changes, I thought that I covered those already, but looks like I missed some spots.

mkalcok:
The documentation was added and the comments from previous round of reviews were addressed. Regarding the docs it was mainly additions, I changed couple of other things I encountered. In the `service-control` docs I also changed:
* emphasis that the `snap` is not recommended way to adjust services
* Titles that used gerunds, because our documentation guide advises against them

This should be now fully prepared for review and merge

mkalcok:
Interesting, a test (ovn_ovs_cli_ipv6) that was [passing fine](https://github.com/canonical/microovn/actions/runs/16319557888/job/46093950890) before, seems to be stuck now after fixing a typo in docs :thinking:
I'll investigate further.

docs/how-to/datapath-only-mode.rst
```diff
@@ -0,0 +1,112 @@
+==================
+Datapath-only Mode
+==================
+
+This is a mode of operation in which MicroOVN cluster runs only ``switch`` and
+``chassis`` service, with ``central`` service being provided externally by another
+MicroOVN cluster or by some other OVN deployment.
```
fnordahl, line 7: It is pertinent to put emphasis on the main use case being migration / transition.

In the current wording a unsuspecting user might mistake this feature to OVN interconnect or similar where two deployments actually can work together, which is not the case for this particular feature.

docs/how-to/datapath-only-mode.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Bootstrap MicroOVN cluster
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First we bootstrap the MicroOVN cluster on the ``node-1`` and generate tokens, so that
+``node-2`` and ``node-3`` can join.
+
+.. code-block::
+
+   $ sudo microovn init
```
fnordahl, line 48: No terminal prefix in code-blocks, make two code-blocks instead, one with the command, another with the output/interaction.

docs/how-to/datapath-only-mode.rst
```diff
@@ -0,0 +1,112 @@
...
+   What's the name of the new MicroOVN server? (empty to exit): node-2
+   <NODE_2_TOKEN>
+   What's the name of the new MicroOVN server? (empty to exit): node-3
+   <NODE_3_TOKEN>
+
+Then we join the cluster with ``node-2`` and ``node-3``
+
+.. code-block::
+
+   $ sudo microovn init
```
fnordahl, line 67: No terminal prefix in code-blocks, make two code-blocks instead, one with the command, another with the output/interaction.

docs/how-to/datapath-only-mode.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Verification
+------------
+
+To verify that the configuration was successfully applied, we can check that our
+chassis successfully registered with the OVN Southbound database.
+
+.. code-block::
+
+   $ sudo microovn.ovn-sbctl show
```
fnordahl, line 93: No terminal prefix in code-blocks, make two code-blocks instead, one with the command, another with the output/interaction.

microovn/ovn/environment/environment.go
fnordahl: nit: thanks for putting it back, it appears to be removed in commit 2 and put back in commit 5, so if you have an opportunity to move it to commit 2 that would give a cleaner git history.

fnordahl (changes requested):

PR #248, opened 2025-07-18
node: Fix incomplete remote addresses in northd

mkalcok:
MicroOVN used incomplete list of NB/SB cluster members when starting the ovn-northd service. When central service was enabled, local ovn.env file, containing list of IP addresses of NB/SB cluster members, was not updated before starting of the northd service. This caused northd service, on a node that just enabled the central service (or joined the cluster), to miss its own IP address.

This issue was partially hidden by the fact that all other northd services in the cluster had complete list of remote IPs. It was only the last joining member that was incomplete.
For the problem to manifest fully, the northd instance with the incomplete list of remote addresses would have to aquire lock, and the NB/SB database that was missing from its list would have to become the cluster leader.

A new test was added to ensure that all northd instances in the cluster have complete list of IPs of the NB/SB cluster members.

Fixes: ad657d813db24fad950171ddd454292f7b9d5966

### DIFF
--- a/microovn/node/node.go
+++ b/microovn/node/node.go
@@ -17,6 +17,7 @@ import (
 	"github.com/canonical/microovn/microovn/ovn/certificates"
 	ovnCluster "github.com/canonical/microovn/microovn/ovn/cluster"
 	ovnCmd "github.com/canonical/microovn/microovn/ovn/cmd"
+	"github.com/canonical/microovn/microovn/ovn/environment"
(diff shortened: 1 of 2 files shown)


mkalcok:
Closing in favor of #249 to get TIOBE run from the main repository.
Main changes from this iteration are adjustments to `OVN_INITIAL_*` variables. See commit message for details.

PR #250, opened 2025-07-22
CI: Use self-hosted runners for TIOBE job

mkalcok:


fnordahl (approved):
LGTM, thank you for taking care of this!

PR #257, opened 2025-08-18
daemon: Change default listening port

mkalcok:
The old default port of MicroOVN daemon (6443) conflicts with the default port of Kubernetes API [0]. Therefore, we are picking a new default of 6686. Brief search did not turn up any application commonly using this port, and it is currently marked as 'Unassigned' in the IANA list [1].

Note this change affects only new installations. Any MicroOVN instances deployed from prior versions will retain the old port binding.

[0] https://kubernetes.io/docs/concepts/security/controlling-access/#transport-security
[1] https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt

### DIFF
--- a/microovn/cmd/microovn/main.go
+++ b/microovn/cmd/microovn/main.go
@@ -15,7 +15,7 @@ import (
 
 // DefaultMicroClusterPort is the default port used for the MicroOVN
 // MicroCluster Daemon.
-const DefaultMicroClusterPort = 6443
+const DefaultMicroClusterPort = 6686
 
 // CmdControl has functions that are common to the microctl commands.
 // command line tools.


mkalcok:
Oddly, the `make spelling` runs fine locally. It looks related to the [recent commit](https://github.com/canonical/documentation-workflows/commit/55b0d3238587f2cf3a4f1d7199c8b15dc71d573b) in the Canonical's documentation GH action that we use. I'll look into it.

mkalcok:
Based on the discussion in the [PR related to the above linked commit](https://github.com/canonical/documentation-workflows/pull/34) It looks like the change is going to be reverted.

fnordahl (approved):
LGTM, thanks for the fix!

I took it for a spin mixing old and new version, and it magically just works. Seems the implementation has some good choices as to where to store/retrieve this information for existing cluster members!

PR #258, opened 2025-08-27
build: Add snap slots for ovn exporter support

jneo8:
Add ovn, ovn-central, ovn-commands for exporter requirement

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -27,6 +27,12 @@ slots:
     source:
       write:
         - "$SNAP_COMMON/run/switch"
+        - "$SNAP_COMMON/run/ovn"
+  ovn-central-data:
+    interface: content
+    source:
+      read:
+        - "$SNAP_COMMON/data/central"
 
 hooks:
   install:


snap/snapcraft.yaml
```diff
@@ -27,6 +27,22 @@ slots:
     source:
       write:
         - "$SNAP_COMMON/run/switch"
+  ovn:
```
mkalcok, line 30: This should be probably named `ovn-central` to match naming convention of already existing `ovn-chassis`, which also exposes runtime directory.

snap/snapcraft.yaml
```diff
@@ -27,6 +27,22 @@ slots:
...
+    interface: content
+    source:
+      write:
+        - "$SNAP_COMMON/run/ovn"
+  ovn-central:
```
mkalcok, line 35: if `ovn` gets renamed to `ovn-central`, this could then be named `ovn-central-data`.

jneo8:
The CI failed because of missing token.

mkalcok (approved):
LGTM, thanks for the update.

PR #259, opened 2025-08-27
Changes microovn bgp to use a veth pair

crypticC0der:
Adds netplan integration, which is so far only used to generate a veth pair, I would like to expand it to include the interface bridge created and also rework how you build a netplan yaml file.

The main change is this avoids remastering the eth-bgp interface to the vrf, we instead have a veth peer with one side on ovn and the other on the vrf.


microovn/frr/bgp/redirect.go
```diff
@@ -288,6 +298,39 @@ func createVrf(ctx context.Context, s state.State, extConnections []types.BgpExt
 	return nil
 }
 
+func generateVeth(ctx context.Context, _ state.State, interfaceName string, macs []string, tableID string) error {
+	var confBuilder strings.Builder
+
+	vethname := getBgpVethName(interfaceName)
+	vrfName := getVrfName(tableID)
+	// Ensure we don't announce any default route from VRF to our peer.
```
fnordahl, line 306: This comment feels out of place?

microovn/frr/bgp/redirect.go
```diff
@@ -288,6 +298,39 @@ func createVrf(ctx context.Context, s state.State, extConnections []types.BgpExt
...
+	}
+
+	fmt.Fprintf(&confBuilder, "    %s-b:\n      peer: %s-a\n", vethname, vethname)
+	if len(macs) >= 2 {
+		fmt.Fprintf(&confBuilder, "      macaddress: %s\n", macs[1])
+	}
+
+	if tableID != "" {
+		fmt.Fprintf(&confBuilder, "  vrfs:\n    %s:\n      table: %s\n      interfaces:\n        - %s-a\n", vrfName, tableID, vethname)
+	}
```
fnordahl, line 323: The go [yaml package](https://pkg.go.dev/gopkg.in/yaml.v3) allows seamless marshaling of structs into valid YAML.

Let's make use that instead of patching strings together.

microovn/frr/bgp/redirect.go
```diff
@@ -327,24 +370,23 @@ func redirectBgp(ctx context.Context, s state.State, extConnections []types.BgpE
...
+		err = generateVeth(ctx, s, extConnection.Iface, []string{mac}, tableID)
+		if err != nil {
+			return err
+		}
+
 		// Create OVS port and associate it with the LSP
 		_, err = ovnCmd.VSCtl(ctx, s,
 			"--",
-			"add-port", intBr, bgpIface,
+			"add-port", intBr, brgIface,
```
fnordahl, line 381: An important deployment tooling design principle is to delegate as much as possible to system tooling, allowing the system to operate independently of microcluster daemon malfunction.

Netplan is capable of adding veths to a Open vSwitch bridge, so lets see if we can make it do that too.

Something like:
```yaml
network:
  version: 2
  bridges:
    br-int:
      interfaces: [ veth0 ]
      openvswitch:
        fail-mode: secure
```

Something like the below would give you a questing container in which the required version of Netplan is available:
```diff
diff --git a/tests/test_helper/setup_teardown/bgp_data_plane.bash b/tests/test_helper/setup_teardown/bgp_data_plane.bash
index 73d25fc..e23e8f7 100644
--- a/tests/test_helper/setup_teardown/bgp_data_plane.bash
+++ b/tests/test_helper/setup_teardown/bgp_data_plane.bash
@@ -17,7 +17,10 @@ setup_file() {
 
     TEST_CONTAINER=$(container_names "$BATS_TEST_FILENAME" 1 | tr -d '[:space:]')
     export TEST_CONTAINER
+    MICROOVN_TEST_CONTAINER_IMAGE=ubuntu-daily:questing
+    export MICROOVN_TEST_CONTAINER_IMAGE
     launch_containers_args "-c linux.kernel_modules=vrf,openvswitch -c security.nesting=true" $TEST_CONTAINER
+    unset MICROOVN_TEST_CONTAINER_IMAGE
     wait_containers_ready $TEST_CONTAINER
     install_microovn "$MICROOVN_SNAP_PATH" $TEST_CONTAINER
     bootstrap_cluster $TEST_CONTAINER
```

Note that I had to pause the test and manually run netplan one more time to make the test work, so there is likely a race/timing issue somewhere.

microovn/frr/bgp/redirect.go
```diff
@@ -137,10 +138,19 @@ func getVrfName(tableID string) string {
 	return fmt.Sprintf("ovnvrf%s", tableID)
 }
 
+func getBgpVethName(externalIface string) string {
+	return fmt.Sprintf("v-%s-bgp", externalIface)
```
fnordahl, line 142: Interface name real estate is scarce:
https://github.com/torvalds/linux/blob/07d9df80082b8d1f37e05658371b087cb6738770/include/uapi/linux/if.h#L33

Can you think of a way to make this shorter?

fnordahl (changes requested):

snap/snapcraft.yaml
```diff
@@ -44,6 +44,7 @@ apps:
       - network
       - network-bind
       - network-control
+      - network-setup-control
```
mkalcok, line 47: With the additional plug, [this doc section](https://github.com/canonical/microovn/blob/d49dc8f1574ec927fcdcf15e783507641d0cdec3/docs/developers/building.rst?plain=1#L87-L96) needs an update as well.

microovn/frr/bgp/redirect.go
```diff
@@ -455,6 +503,33 @@ func teardownAll(ctx context.Context, s state.State) error {
...
+				continue
+			}
+			_, err = shared.RunCommandContext(ctx, "ip", "link", "delete", "dev", iface)
+			if err != nil {
+				allErrors = errors.Join(allErrors, fmt.Errorf("failed remove interface '%s': %v", iface, err))
+			} else {
+				deletedPeers[ifaceData.Peer] = true
+			}
+		}
+	}
```
fnordahl, line 526: What about the VRFs, do you need to remove those too?

fnordahl:
```diff
diff --git a/microovn/frr/bgp/redirect.go b/microovn/frr/bgp/redirect.go
index 53850e6..f4e8d2c 100644
--- a/microovn/frr/bgp/redirect.go
+++ b/microovn/frr/bgp/redirect.go
@@ -524,6 +524,13 @@ func teardownAll(ctx context.Context, s state.State) error {
                        }
                }
        }
+       _, _ = shared.RunCommandContext(ctx, "ip", "route", "flush", "table", "10")
+       _, _ = shared.RunCommandContext(ctx, "ip", "route", "flush", "table", "20")
+       _, _ = shared.RunCommandContext(ctx, "ip", "route", "flush", "table", "30")
+       _, _ = shared.RunCommandContext(ctx, "ip", "link", "delete", "dev", "ovnvrf10")
+       _, _ = shared.RunCommandContext(ctx, "ip", "link", "delete", "dev", "ovnvrf20")
+       _, _ = shared.RunCommandContext(ctx, "ip", "link", "delete", "dev", "ovnvrf30")
+       netplan.Apply(ctx)
 
        _, err = shared.RunCommandContext(ctx, "rm", "/tmp/90-microovn-bgp-veth.yaml")
        if err != nil {
```

@crypticC0der Instrumenting the code with ^ is doing something, and appears to make the control plane tests pass.

fnordahl (approved):
Thank you for your thorough work on this and for the added netplan package with test coverage!

In the interest of getting us to the next challenge, I took the liberty of adding a small commit  that tightens the file mode for netplan yaml fragments.

PR #260, opened 2025-08-29
lint: Skip package name check.

crypticC0der:
'revive' started to warn against "meaningless package names"[0] and it gets triggered by our "types" package.
I don't presume to know more about good practices than Go developers, but I struggle to come up with a good alternative to our "types" package. I believe that in the context of the project structure, package 'microovn.api.types' makes sense.

[0] https://go.dev/blog/package-names#bad-package-names


fnordahl (approved):
LGTM, thanks!

PR #261, opened 2025-09-02
CI: Run TIOBE job only in Canonical repo

mkalcok:
With TIOBE job running on self-hosted runners,
the job never finishes on forks that are outside
the Canonical org. The whole pipeline is held-up
for 24h until the job times out while waiting for
the runner.


fnordahl (approved):
LGTM, thanks!

PR #264, opened 2025-09-08
Rebase branch experimental/bgp ontop of main

crypticC0der:
Rebases the branch experimental/bgp ontop of main, and handles the conflicts this causes

### DIFF
--- a/docs/contributors/code.rst
+++ b/docs/contributors/code.rst
@@ -100,6 +100,7 @@ do it automatically for locally installed snaps.
                 hardware-observe \
                 hugepages-control \
                 network-control \
+                network-setup-control \
                 openvswitch-support \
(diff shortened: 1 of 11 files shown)


fnordahl (changes requested):
Thanks for your work on this!

There are a few of these commits that can be omitted, for example the commits that add patches only to later remove them can likely be removed all together.

There are also some examples of experimental churn that can be squashed so that only the final solution remains, one example being the iBGP to eBGP change.

Commits with WIP likely need to be redone, one example being the IPv6 RA for LRPs. I think the temporary choice of 1 second is too aggressive.

Finally, now that a core26 base snap is available in the edge channel, we likely want to use binaries from there rather than building from source.

PR #270, opened 2025-10-02
Adds a note about needed aliasing with netplan

crypticC0der:
Netplan requires at least the ovs-vsctl alias, this is currently undocumented, which is a problem. This adds a simple note in the aliasing section to explain this.


fnordahl (approved):
LGTM, thanks!

PR #271, opened 2025-10-02
BGP: Replace FRR with Bird

mkalcok:
This is a quite large change, but I believe it's best that the switchover happens in one atomic commit. The reason it's still a WIP is in the "bgp_dataplane" tests. There's a sleep required after enabling BGP, otherwise the ping at the end of the test wont pass. More info in the inline comment. Do note that even without the sleep, other BGP-related functionality works without issues (BGP session is established, routes are exchanged, ...). It's only the actual flow of data, that's somehow affected.

If you have time to help me look into this @fnordahl / @crypticC0der I'd appreciate any pointers.


mkalcok:
Thanks for the great find @fnordahl. I kept the test helper for Mac_Binding in the separate commit and squashed the fix for the dataplane test (along with the strongly couple `microovn_bgp_neighbor_address` helper) into the big Bird commit. You were also added as Co-author for correct attribution.

I think that the workaround in the tests is sufficient from the POV of MicroOVN. The fact that OVN needs to do Neighbor Discovery on first reply is just a fact of life at this point and not something that MicroOVN can do much about.

I think that in "regular" IPv4 only traffic, the MAC_Binding would be populated by the incoming traffic. However in this scenario, where next-hop of the returning traffic is IPv6 address, we don't have that luxury. We could propose an upstream patch for ovn-controller to proactively try to do the ND when it learns IPv4 route with link-local IPv6 next-hop. But that a topic for another day.

snapcraft/bird.env
```diff
@@ -0,0 +1,5 @@
+export BIRD_RUN_DIR="${SNAP_COMMON}/run/bird"
+export BIRD_DATA_DIR="${SNAP_COMMON}/data/bird"
+export BIRD_PID="$BIRD_RUN_DIR/bird.pid"
+export BIRD_SOCKET="$BIRD_RUN_DIR/bird.ctl"
+export BIRD_CONFIG="$BIRD_DATA_DIR/bird.conf"
```
fnordahl, line 5: Newline at end of file, please.

snap/snapcraft.yaml
```diff
@@ -213,130 +202,24 @@ parts:
...
-      - libacl1
-      - libattr1
-      - libaudit1
-      - libselinux1
-      - libxtables12
-    plugin: autotools
-    source: https://github.com/FRRouting/frr.git
-    source-type: git
-    source-tag: frr-10.1.1
-    autotools-configure-parameters:
```
fnordahl, line 299: Let's keep path configuration that would be compatible with switching to the .deb when we're ready for that. That would also get rid of these long `usr/local/` paths.

https://salsa.debian.org/debian/bird3/-/blob/debian/experimental/debian/rules?ref_type=heads#L12-L15

snap/hooks/install
```diff
@@ -19,11 +19,10 @@ EOF
...
 
-# Create directories required by FRR
-frr_etc="$SNAP_COMMON/data/frr/etc/"
-mkdir -p $frr_etc
-mkdir -p "$SNAP_COMMON/data/frr/tmp"
-mkdir -p "$SNAP_COMMON/run/frr"
+# Create directories required by Bird
+bird_data_dir="$SNAP_COMMON/data/bird"
+mkdir -p "$bird_data_dir"
+mkdir -p "$SNAP_COMMON/run/bird"
```
fnordahl, line 25: The other directories in `$SNAP_COMMON/data` and `$SNAP_COMMON/run` have `0o700` permissions, while these are created with `0o755`.

I don't see the other scripts being explicit about it, so it may be install hook is ran with a different umask.

In any case, let's be consistent and tighten the permissions for these directories.

PR #273, opened 2025-10-08
Add acceptRa option to veth pair

crypticC0der:
Allows setting acceptRa to false in a veth pair to stop systemd-networkd processing the ipv6 RA packets.


fnordahl (changes requested):
Overall the approach looks good to me, but I'd like one more test case added.

The goal of this change is to ensure the IPv6 default route in the VRF does not come from the RA routing protocol, and in a subsequent PR/task for this pulse we'll ensure IPv6 default route can be learned from BGP.

The IPv6 router advertisement processing behavior of Linux is currently a symbiose of the Linux kernel and systemd-networkd, and it would be important for us to make sure it behaves the way we expect it to.

If we at a later date get the route both from RA and BGP we might get into trouble.

So the test would be to ensure we don't have a IPv6 default route from routing protocol RA in the VRF.

fnordahl (approved):
LGTM, thanks!

PR #322, opened 2026-03-31
Fix broken switchstart test in main.

crypticC0der:
I have no idea how this test is passing in main, genuinely


matperin (approved):
I encountered the same problem on my currently up PR.
I have tested it locally and the fix indeed solves the switchstart suite failure.

PR #328, opened 2026-04-14
Minor dpu reworks.

crypticC0der:
- lspci is inaccessable by the snap outside of --dangerous, so we fix this
- we need a slight rework to the externalIDs logic when nothing else exists there.

### DIFF
--- a/.github/workflows/build-microovn.yaml
+++ b/.github/workflows/build-microovn.yaml
@@ -46,7 +46,7 @@ jobs:
           sudo snap set lxd daemon.group=adm
           sudo lxd init --auto
           if [ "${POSSIBLE_TARGET_BRANCH}" = "main" ]; then
-            export SNAPCRAFT_CHANNEL=latest/edge
+            export SNAPCRAFT_CHANNEL=latest/candidate # latest edge is currently broken, so we have to use candidate for now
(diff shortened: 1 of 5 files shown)


matperin (changes requested):
Maybe it would be better to split [Minor dpu reworks.](https://github.com/canonical/microovn/pull/328/changes/526c0457c86a12fd5ff109893bd78503945b2b5f) into two commits?
I know that they are very minor changes but I think they achieve different things and it is better to have separate commit messages for future reference.

Other than that the rest seems good. Do we know why snapcraft latest/edge is not pulling dependencies properly, though? This is probably something we should fix at the root, if we have the chance.

crypticC0der:
> Do we know why snapcraft latest/edge is not pulling dependencies properly, though?

the version in latest edge (post104) only pulls in explicitly mentioned packages and not their dependancies, which is a pain and I am expecting this to be a bug that gets reverted before it hits stable, i have put a message in their channel asking about it.

matperin (dismissed):
Thanks for the change, looks good!

matperin (approved):
The added `snap refresh` for `snapd` inside the container seems a good addition, I will confirm my previous approval.

PR #332, opened 2026-04-30
Mark BGP as experimental.

crypticC0der:
Updates the documentation to ensure the user is aware that this feature is not finalized and is heavily subject to change, and therefore deployments using MicroOVN BGP in its current state will not be supported.

Also adds a note when you enable the bgp service that you are using experimental functionality.

### DIFF
--- a/.github/workflows/build-microovn.yaml
+++ b/.github/workflows/build-microovn.yaml
@@ -46,7 +46,7 @@ jobs:
           sudo snap set lxd daemon.group=adm
           sudo lxd init --auto
           if [ "${POSSIBLE_TARGET_BRANCH}" = "main" ]; then
-            export SNAPCRAFT_CHANNEL=latest/candidate # latest edge is currently broken, so we have to use candidate for now
+            export SNAPCRAFT_CHANNEL=latest/stable # latest edge is currently broken, so we have to use stable for now
           fi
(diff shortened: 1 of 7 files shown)


crypticC0der:
There is an interesting bug with bgp as is present wherin the peer responds to the router advertisement with `16:24:19.610076 00:16:3e:97:1d:8e > 02:90:b1:a2:e6:28, ethertype IPv6 (0x86dd), length 94: (class 0xc0, flowlabel 0xe2046, hlim 1, next-header TCP (6), payload length 40) fe80::216:3eff:fe97:1d8e.48278 > fe80::90:b1ff:fea2:e628.179: Flags [S], cksum 0xf1c6 (incorrect -> 0xce8d), seq 1318403995, win 64800, options [mss 1440,sackOK,TS val 3192488198 ecr 0,nop,wscale 10], length 0` the notable thing being the incorrect checksum which is causing it to discarded by microovn-bgp-data-plane-1 and therin the connection fails to establish, this seems to be only affecting us now as ubuntu:lts is now 26.04

I will investigate this more

matperin (dismissed):
Seems good to me! Thank you for adding this specification to the docs, it will avoid confusion among the users with the future BGP development.

matperin (approved):

PR #333, opened 2026-05-06
tests: Fix BGP peering failures

matperin:
Fix regressions that prevent BGP sessions from establishing in the MicroOVN BGP redirect feature:

1. netplan: Add link-local configuration for BGP veth interfaces

   On newer systemd-networkd, veth interfaces require explicit
   link-local: [ipv6] configuration to receive an IPv6 link-local
   address, which is required for unnumbered BGP peering.

2. bgp/redirect: Work around networkd not fully configuring interfaces

   After netplan apply, explicitly ensure BGP-side veths are:
   - Bound to the VRF (networkd may not apply this)
   - Brought UP (networkd may leave them DOWN)

3. tests: Disable FRR AppArmor profiles in test containers

   Newer FRR package versions ships AppArmor profiles that confine
   bgpd in enforce mode. The bgpd profile causes EACCES on socket
   read() for BGP connections that traverse OVS/OVN virtual network
   paths (veth -> OVS bridge -> patch port -> OVN LRP -> veth in VRF).
   Disable these profiles after FRR installation in test containers.

### DIFF
--- a/microovn/bgp/redirect.go
+++ b/microovn/bgp/redirect.go
@@ -446,8 +446,11 @@ func generateVeth(ctx context.Context, s state.State, extConnections []types.Bgp
 		mac := generateLrpMac(getLrpName(s, extConnection.Iface))
 
 		// Add to virtual ethernet
-		np.AddVeth(bgpInterface, brgInterface, mac, false)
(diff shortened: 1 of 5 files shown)


crypticC0der (approved):
this looks great thanks for sorting these out

PR #335, opened 2026-05-12
Automated release script.

crypticC0der:
A script to automate the steps required to release stable MicroOVN versions.

Assisted-by: Gemini

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,287 @@
+#!/bin/bash
+set -e
+
+DELETE_EXISTING=0
+SKIP_TESTS=0
+
+usage() {
+    echo "Usage: $0 [-d] [-h] [-s] [VERSION]"
+    echo ""
+    echo "Automates the MicroOVN release process."
+    echo ""
+    echo "Options:"
+    echo "  -d        Delete existing PR and stable branches before running"
+    echo "  -h        Show this help message and exit"
+    echo "  -s        Skip upgrade tests"
+    echo ""
+    echo "Arguments:"
+    echo "  VERSION   Optional. The release version (e.g., 24.03)."
+    echo "            Defaults to the current year with .03 (e.g., $(date +%y.03))."
+}
+
+test_setup_file() {
+    echo "tests/test_helper/setup_teardown/$(basename "${1//.bats/.bash}")"
+}
+
+# Create test file content
+create_test_content() {
+    ln -s upgrade.bats "$1"
+    ln -s upgrade.bash "$(test_setup_file "$1")"
+}
+
+# Parse options
+while getopts "dhs" opt; do
+  case $opt in
+    d)
+      DELETE_EXISTING=1
+      ;;
+    h)
+      usage
+      exit 0
+      ;;
+    s)
+      SKIP_TESTS=1
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      usage
+      exit 1
+      ;;
+  esac
+done
+
+shift $((OPTIND - 1))
+
+# Handle optional version argument
+if [ -n "$1" ]; then
+    VERSION="$1"
+else
+    VERSION="$(date +%y.03)"
+fi
+
+PR_BRANCH="prepare-$VERSION"
+STABLE_BRANCH="branch-$VERSION"
+YAML_FILE="snap/snapcraft.yaml"
+
+# Ensure we have the target file
+if [ ! -f "$YAML_FILE" ]; then
+    echo "Error: $YAML_FILE not found! Please run this script from the repository root."
+    exit 1
+fi
+
+if [ "$DELETE_EXISTING" -eq 1 ]; then
+    echo "-> [-d] Removing existing branches if they exist..."
+    git branch -D "$PR_BRANCH" 2>/dev/null || true
+    git branch -D "$STABLE_BRANCH" 2>/dev/null || true
+    echo
+fi
+
+if git branch | grep -qE "($STABLE_BRANCH|$PR_BRANCH)"; then
+    if git branch | grep -q "$STABLE_BRANCH"; then
+        echo "Error: $STABLE_BRANCH already exists"
+    fi
+
+    if git branch | grep -q "$PR_BRANCH"; then
+        echo "Error: $PR_BRANCH already exists"
+    fi
+    exit 1
+fi
+
+
+
+echo "Setting up for MicroOVN release:"
+
+echo "-> Checking out 'main'..."
+git checkout main
+
+# 2. Create the working branch for the PR
+echo "-> Creating PR branch: $PR_BRANCH"
+git checkout -b "$PR_BRANCH"
+
+echo
+echo "Stabilising the branch:"
+
+# Delete any build-base and any grade statement
+sed -i -E "/^build-base:.*/d" "$YAML_FILE"
+sed -i -E "s/^grade: devel/grade: stable/" "$YAML_FILE"
+echo "-> Auto-pinning git sources to their latest stable tags..."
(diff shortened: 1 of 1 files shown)


scripts/release.sh
fnordahl, line 1: Should we find a suitable directory to keep this file in?

Do we want to lint it with shellcheck?

scripts/release.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# 2. Create the working branch for the PR
+echo "-> Creating PR branch: $PR_BRANCH"
+git checkout -b "$PR_BRANCH"
+
+echo
+echo "Stabilising the branch:"
+
+# Delete any build-base statement
+sed -i -E "/^build-base:.*/d" "$YAML_FILE"
```
fnordahl, line 105: Should we also get rid of the `grade` 'devel' line?

scripts/release.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+# Validate arguments
+
+VERSION="$(date +%y.03)"
```
fnordahl: This assumes we only ever release at a LTS cadence, which so far appears to be true

fnordahl (changes requested):
I love scripting of manual tasks, looks like it was quite effectively done with the help from an LLM as well, which is welcome!

I have some requests in-line.

scripts/release.sh
crypticC0der: added version arg

scripts/release.sh
crypticC0der: shellcheck is happy with it, I dont see why we wouldnt

mkalcok (commented):
I'm very much +1 on scripting these release tasks, thanks.
One thing that is not addressed yet are the upgrade tests. They take argument that controls "from which versions" is the upgrade test performed. I'd say that the new release branch should run test from previous release (and latest previous LTS?), while the new 'main' should run upgrade from the new release branch.

It also may be worth to throw in a doc page that describes the release process. 

Sorry for the shallow review, I'm on the mobile and just wanted to quickly mention the upgrade thing.

Regarding the scripting language choice, LLMs skew heavily towards shell and grep/sed/awk any chance they get. While not necessarily wrong, imo yaml could be much more cleanly manipulated programmatically either from python or using yq. However that's not a big deal from my side.

crypticC0der:
> I'm very much +1 on scripting these release tasks, thanks.

Thank you

>  One thing that is not addressed yet are the upgrade tests. They take argument that controls "from which versions" is the upgrade test performed. I'd say that the new release branch should run test from previous release (and latest previous LTS?), while the new 'main' should run upgrade from the new release branch.

Good shout :+1: 

> It also may be worth to throw in a doc page that describes the release process.

https://github.com/canonical/microovn/blob/main/docs/reference/release-process.rst
We could expand this to mention the upgrade tests, but this does exist

> Regarding the scripting language choice, LLMs skew heavily towards shell and grep/sed/awk any chance they get. While not necessarily wrong, imo yaml could be much more cleanly manipulated programmatically either from python or using yq. However that's not a big deal from my side.

To be fair so do I, I think making this a python script would be overkill and it would make parts such as getting the latest tag from the git repo slightly messier. The only real part python could do better than bash is the yaml editing, which is quite a small part of the script.

Anyway thanks for the review, it is always nice to hear from you, we are missing you in spain.

mkalcok:
> > I'm very much +1 on scripting these release tasks, thanks.
> 
> Thank you
> 
> > One thing that is not addressed yet are the upgrade tests. They take argument that controls "from which versions" is the upgrade test performed. I'd say that the new release branch should run test from previous release (and latest previous LTS?), while the new 'main' should run upgrade from the new release branch.
> 
> Good shout 👍
> 

cheers :)

> > It also may be worth to throw in a doc page that describes the release process.
> 
> https://github.com/canonical/microovn/blob/main/docs/reference/release-process.rst We could expand this to mention the upgrade tests, but this does exist
> 

Ah, I did have a faint memory of such page, but I was in a hurry. Anyway, yeah, mentioning the existence of this script on that page makes sense.

> > Regarding the scripting language choice, LLMs skew heavily towards shell and grep/sed/awk any chance they get. While not necessarily wrong, imo yaml could be much more cleanly manipulated programmatically either from python or using yq. However that's not a big deal from my side.
> 
> To be fair so do I, I think making this a python script would be overkill and it would make parts such as getting the latest tag from the git repo slightly messier. The only real part python could do better than bash is the yaml editing, which is quite a small part of the script.
> 

fair enough.
> Anyway thanks for the review, it is always nice to hear from you, we are missing you in spain.

Ah! It's that time of the year already. Enjoy the Spain an please relay my best wishes to the rest of the team :)

scripts/release.sh
```diff
@@ -0,0 +1,301 @@
...
+    echo "  -h        Show this help message and exit"
+    echo "  -s        Skip upgrade tests"
+    echo ""
+    echo "Arguments:"
+    echo "  VERSION   Optional. The release version (e.g., 24.03)."
+    echo "            Defaults to the current year with .03 (e.g., $(date +%y.03))."
+}
+
+# Create test file content
+create_test_content() {
```
fnordahl, line 27: It appears this function basically cargo cults the contents of `test_helper/bats/upgrade.bats`.

Looking at the upgrade setup/teardown code and the `get_upgrade_test_version` function:
https://github.com/canonical/microovn/blob/1ed376b544cc4425f6507c792dfd60831383f0ab/tests/test_helper/setup_teardown/upgrade.bash#L17-L19
https://github.com/canonical/microovn/blob/1ed376b544cc4425f6507c792dfd60831383f0ab/tests/test_helper/common.bash#L138-L175

It appears to me it is the filename that bears the crucial information, the contents of the file remains the same.

Could we perhaps just create a symlink with the appropriate name pointing to `test_helper/bats/upgrade.bats`?

scripts/release.sh
```diff
@@ -0,0 +1,305 @@
...
+# 2. Create the working branch for the PR
+echo "-> Creating PR branch: $PR_BRANCH"
+git checkout -b "$PR_BRANCH"
+
+echo
+echo "Stabilising the branch:"
+
+# Delete any build-base and any grade statement
+sed -i -E "/^build-base:.*/d" "$YAML_FILE"
+sed -i -E "/^grade: devel/d" "$YAML_FILE"
```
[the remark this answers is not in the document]
fnordahl: I tend to agree with Copilot's analysis here, looking at https://documentation.ubuntu.com/snapcraft/stable/reference/snapcraft-yaml/index.html#grade it appears we would need `grade: stable` in order to have the snap published.

I have no comment on its proposed solution.

scripts/release.sh
```diff
@@ -0,0 +1,301 @@
...
+        read -r response
+        if [[ ! "$response" =~ ^([yY][eE][sS]|[yY])$ ]]; then
+            echo "Aborting release process."
+            exit 1
+        fi
+        echo "Continuing despite test failures..."
+    fi
+else
+    echo
+    echo "-> Skipping upgrade tests (-s flag provided)"
```
fnordahl, line 234: I like the enforcement of upgrade tests either being run, or revealing that they were skipped. Let's say we agree on a process where we enforce that the release manager, or robots, logs the output of this command as part of the process.  This allows for transparency and accountability!

👍

fnordahl (changes requested):

scripts/release.sh
crypticC0der: :+1:

fnordahl (dismissed):
LGTM, thanks!

crypticC0der:
@matperin

matperin (approved):
Thank you for the improvements to the test setup file path handling since last revision.
Looks good!

PR #336, opened 2026-05-13
Only runs tests if relevant files have changed

crypticC0der:
We have the luxuary of a large and comprehensive test suite, however if one is just modifiying documentation we do not need to run it. By making the changes conditional we will likely save a lot of compute which is good. We should still build and run a few basic tests to ensure things arent particularly broken from external factors though.

I intentionally left the documentation checks non-conditional as they are fast and dont use a lot of compute.

Assisted-by: Deepseek 3.2

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,18 +71,47 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Check for changes to the snap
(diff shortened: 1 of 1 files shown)


matperin (approved):
Limiting the test runs when pushing non code-changing commits is a great idea, thank you for the change!
The changes to the CI workflow seem very reasonable as well.

PR #339, opened 2026-06-19
Bgpbrg

crypticC0der:
Expands microovn enable bgp to accept a bridge as the argument instead of a network interface.


crypticC0der:
Currently a failed join means that node cannot be added to the cluster, however there is an open issue and PR https://github.com/canonical/microcluster/issues/784

microovn/api/types/setup_bgp.go
```diff
@@ -0,0 +1,38 @@
...
+
+const BgpConfigKeyInterface = "microovn-bgp-arg-interface"
+
+const BgpConfigKeyBridge = "microovn-bgp-arg-br"
+
+const BgpConfigKeyAsn = "microovn-bgp-arg-asn"
+
+const BgpConfigKeyAsnRange = "microovn-bgp-arg-asn_range"
+
+const BgpConfigKeyConnectionIP4Range = "microovn-bgp-arg-connection-ip4-range"
```
matperin, line 38: Maybe it is better to wait for the numbered BGP implementation before introducing the `--connection-ipv4-range` configuration value?
Right now this key seems to be never used and its addition would semantically belong to the proper commit making use of it.

microovn/bgp/redirect.go
```diff
@@ -440,10 +471,10 @@ func generateVeth(ctx context.Context, s state.State, extConnections []types.Bgp
 	vrfInterfaces := []string{}
 	var drPortMapping strings.Builder
 
-	for _, extConnection := range extConnections {
-		bgpInterface := getBgpRedirectIfaceName(extConnection.Iface)
-		brgInterface := getBgpRedirectIfacePeerName(extConnection.Iface)
-		mac := generateLrpMac(getLrpName(s, extConnection.Iface))
+	for _, brg := range bridges {
+		bgpInterface := getBgpRedirectIfaceName(brg.Bridge)
```
matperin, line 475: It is worth nothing that net iface names are capped at 15 chars in the linux kernel.
Since veth names are derived as `v<bridge>-bgp` / `v<bridge>-brg` that would make only bridges with names of <=10 chars acceptable.
Not a problem with the implementation per-se as it is a hard limit on the kernel, but something we may want to document or flag somewhere.

microovn/bgp/service.go
```diff
@@ -149,22 +153,36 @@ func EnableService(ctx context.Context, s state.State, extraConfig *types.ExtraB
 		logging.Debugf("Auto-selected VRF table ID: %s", vrfTableID)
 	}
 
-	err = createExternalBridges(ctx, s, extConnections)
+	cfgBridges, err := extraConfig.ParseBridge()
 	if err != nil {
-		return errors.Join(err, DisableService(ctx, s))
+		logging.Errorf("Failed to parse bridges: %v", err)
```
matperin, line 158: Why are we falling through here instead of returning the error?
That would basically mean we are silently treating bridge parse failures as "no bridge given, use external-connection path", is it really what we want?
If so, maybe a comment or an extension  of the error message would be useful for debugging/understanding the code path.

microovn/api/types/services.go
```diff
@@ -143,9 +144,15 @@ type BgpExternalConnection struct {
 	Iface string
 }
 
-// parseAsnRange parses an ASN range string in format "min-max".
+// BgpExternalConnection represents a parsed structure from ExtraBgpConfig.ExternalConnection string.
```
matperin, line 147: Shouldn't this read `BgpBridge` and `ExtraBgpConfig.Bridge`?

matperin (changes requested):
Overall amazing work @crypticC0der!
Also good job on the test coverage, I appreciate it.
I left some minor things to consider in some comments, but overall the PR seems to be in very good shape already.

matperin (approved):

PR #341, opened 2026-06-23
Doubles the vm wait timeout.

crypticC0der:
Previously this was 30, which was causing the tls_cluster test to be quite inconsistent.


crypticC0der:
@matperin

matperin (approved):

PR #353, opened 2026-07-14
Allow MicroOVN init to take values from stdin.

crypticC0der:
The microovn init functionality is not very machine usable, which is a problem because we like machines and the operator charm could gain a lot of extra functionality through the init options.

The simple solution is to allow passing values though init as a yaml object.

### DIFF
--- a/microovn/cmd/microovn/init.go
+++ b/microovn/cmd/microovn/init.go
@@ -17,8 +17,111 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type InitValues struct {
+	Mode                  string   `yaml:"mode,omitempty"`
+	Name                  string   `yaml:"name,omitempty"`
+	Address               string   `yaml:"address,omitempty"`
+	Services              string   `yaml:"services,omitempty"`
+	Token                 string   `yaml:"token,omitempty"`
+	CustomEncapsulationIP string   `yaml:"custom_encapsulation_ip,omitempty"`
+	CustomCACert          string   `yaml:"custom_ca_cert,omitempty"`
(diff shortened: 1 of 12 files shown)


matperin (changes requested):
Thanks @crypticC0der for the changes.

There is nothing worth flagging in what is currently present, the implementation is very well done, imho.

I think there are a few things missing before calling this feature complete, though:
- While the main functionality testing is covered (end-to-end bootstrap + join of a 3-node cluster via preseed YAML) we are still missing coverage for some added paths.
I think adding one preseed test that sets `services` and `custom_encapsulation_ip` (and, ideally, a `custom_ca` variant mirroring `init_cluster_user_ca.bats`) and one test using `additional_servers` might be worthwhile. Also, while not entirely needed, one could also argue to add a couple of negative validation tests (invalid mode / bad YAML / join without token).
- Since this is a new user/machine-facing interface, I would advise against leaving it undocumented. Leaving no mentions of preseed or the YAML fields on the docs risks making the feature undiscoverable by the charm authors who are the stated audience (mostly us, but still good for future-proofing).
I'd suggest a new how-to (e.g. `docs/how-to/preseed-init.rst`) that documents the full `InitValues` YAML schema and which fields apply to bootstrap vs join (i.e. the validation rules, token required for join and forbidden for bootstrap are non-obvious and should be spelled out), the `additional_servers` behavior (issues join tokens, printed to stdout) and maybe a note that if stdin is a TTY the command stays interactive. Also, a bootstrap example and a join example piping YAML into `microovn init` could be useful.

Other than this everything looks good.

matperin (approved):
Thank you for the added tests.
I see that the documentation is still missing but we can add it in a later commit, if needed.

PR #355, opened 2026-09-01
feat: `microovn inspect` command 

Deezzir:
# Overview

* Adds `microovn inspect`, a read-only command for inspecting the health of a MicroOVN deployment.
* Collects cluster topology, service runtime state, OVN database communication and convergence, and environment consistency.
* Supports human-readable and JSON output, including partial results when members or data sources are unavailable.
* Produces authoritative cluster-wide results on voting members. Standby and spare members provide local observations and mark checks requiring cluster authority as `UNKNOWN`.
* Reports `PASS`, `WARNING`, `FAIL`, and `UNKNOWN` results without exposing raw environment values.

## QA

The command was validated with:

* Go unit and race tests
* Functional `inspect.bats` suite

### Single node cluster

image

### Triple node cluster

image

### DIFF
--- a/microovn/api/inspect.go
+++ b/microovn/api/inspect.go
@@ -1,0 +1,80 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/canonical/microcluster/v3/microcluster/rest"
+	"github.com/canonical/microcluster/v3/microcluster/rest/response"
+	"github.com/canonical/microcluster/v3/state"
+	"github.com/canonical/microovn/microovn/api/types"
+	"github.com/canonical/microovn/microovn/inspect/facts"
+	"github.com/canonical/microovn/microovn/ovn/environment"
+	"github.com/canonical/microovn/microovn/snap"
+)
+
+var InspectionSnapshotEndpoint = rest.Endpoint{
+	Path: "inspect/snapshot",
+	Get: rest.EndpointAction{
+		Handler:        inspectionSnapshotGet,
+		AllowUntrusted: false,
+		ProxyTarget:    false,
+	},
+}
+
+var InspectionDatabaseEndpoint = rest.Endpoint{
+	Path: "inspect/database",
+	Get: rest.EndpointAction{
+		Handler:        inspectionDatabaseGet,
+		AllowUntrusted: false,
+		ProxyTarget:    false,
+	},
+}
+
+var InspectionNetworkEndpoint = rest.Endpoint{
+	Path: "inspect/network",
+	Get: rest.EndpointAction{
+		Handler:        inspectionNetworkGet,
+		AllowUntrusted: false,
+		ProxyTarget:    false,
+	},
+}
+
+func inspectionSnapshotGet(s state.State, r *http.Request) response.Response {
+	snapshot := facts.CollectNodeSnapshot(
+		r.Context(),
+		s.Name(),
+		s.Address().Hostname(),
+		snap.Services,
+		environment.ReadEnvironment,
+	)
+	return response.SyncResponse(true, snapshot)
+}
+
+func inspectionDatabaseGet(s state.State, r *http.Request) response.Response {
+	scope := r.URL.Query().Get("scope")
+	if scope == "" {
+		return response.BadRequest(fmt.Errorf("missing required query parameter: scope"))
+	}
+	if scope != string(types.InspectionScopeCluster) && scope != string(types.InspectionScopeLocal) {
+		return response.BadRequest(fmt.Errorf("invalid scope: %s", scope))
+	}
+	probe, err := facts.CollectDatabaseProbe(
+		r.Context(),
+		s,
+		types.InspectionScope(scope),
+	)
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	return response.SyncResponse(true, probe)
+}
+
+func inspectionNetworkGet(s state.State, r *http.Request) response.Response {
+	probe := facts.CollectNetworkProbe(
+		r.Context(),
+		s,
+	)
+	return response.SyncResponse(true, probe)
+}
--- a/microovn/inspect/collect.go
+++ b/microovn/inspect/collect.go
@@ -1,0 +1,422 @@
+package inspect
+
+import (
+	"cmp"
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"net/netip"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/canonical/microcluster/v3/microcluster"
+	microTypes "github.com/canonical/microcluster/v3/microcluster/types"
+	"github.com/canonical/microovn/microovn/api/types"
+	"github.com/canonical/microovn/microovn/client"
+	"github.com/canonical/microovn/microovn/ovn/cmd"
+)
+
+func collectInput(ctx context.Context, cluster *microcluster.MicroCluster, local microTypes.Client) (Input, error) {
+	input := Input{
+		Snapshots: make(map[string]types.InspectionNodeSnapshot),
+		Schemas:   make(map[string]types.InspectionSchemaEvidence),
+	}
+
+	execution, err := collectExecutionContext(ctx, cluster)
+	if err != nil {
+		return Input{}, err
+	}
+	input.ExecutionContext = execution
+
+	snapshot, err := client.GetNodeSnapshot(ctx, local)
+	if err != nil {
+		input.CollectionErrors = append(input.CollectionErrors, types.InspectionCollectionError{
+			Node:      input.ExecutionContext.LocalNode,
+			FactGroup: "snapshot",
+			Message:   fmt.Sprintf("failed to get local node snapshot: %v", err),
+		})
+	} else {
+		snapshot.NodeName = input.ExecutionContext.LocalNode
+		input.Snapshots[snapshot.NodeName] = snapshot
+	}
+
+	if input.ExecutionContext.Authoritative {
+		members, err := cluster.GetClusterMembers(ctx)
+		if err != nil {
+			input.CollectionErrors = append(input.CollectionErrors, types.InspectionCollectionError{
+				Node:      input.ExecutionContext.LocalNode,
+				FactGroup: "member",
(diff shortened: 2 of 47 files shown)


microovn/api/inspect.go
```diff
@@ -0,0 +1,492 @@
...
+
+	switch scope {
+	case types.InspectionScopeCluster:
+		nbCtl = cmd.NBCtlCluster
+		sbCtl = cmd.SBCtlCluster
+	case types.InspectionScopeLocal:
+		nbCtl = cmd.NBCtl
+		sbCtl = cmd.SBCtl
+	default:
+		panic(fmt.Sprintf("invalid scope: %s", scope))
```
matperin: nit: I am a little bit wary about leaving a `panic` here. Even if is unreachable today (the handler validates first), we rely on every future caller making the same validation.
I would suggest:
- returning an error instead of crashing.
- split the function into into two constructors (e.g. `dbCfgReadersCluster` and `dbCfgReadersLocal`) which would avoid the switch.

microovn/inspect/collect.go
```diff
@@ -0,0 +1,404 @@
...
+	if len(matches) != 1 {
+		return types.InspectionExecutionContext{}, fmt.Errorf(
+			"expected one local dqlite member, found %d",
+			len(matches),
+		)
+	}
+
+	role := matches[0].Role
+	name := matches[0].Name
+	if role == "stand-by" {
```
matperin: nit: I would add a brief comment explaining that we are normalizing the dqlite reported role to our internal value.
Also, it may be cleaner to fold the normalization into the switch.

microovn/inspect/service_runtime.go
```diff
@@ -0,0 +1,313 @@
...
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/canonical/microovn/microovn/api/types"
+)
+
+type ServiceRuntimeCheck struct{}
+
+var roleDaemons = map[types.SrvName][]string{
```
matperin: Do we really need a map here?
I ask because at first glance I only see `roleDaemons[role]` only being looked up inside the loop over all roles and never by arbitrary key.
If that is the case, we could just use a ordered `struct` to avoid having `serviceRoles` purely to iterate `roleDaemons` in a deterministic order (which is a maintenance burden).

microovn/inspect/environment.go
```diff
@@ -0,0 +1,304 @@
...
+	environment []types.InspectionEnvironment
+}
+
+type DiffResult struct {
+	Added          []types.InspectionEnvironment
+	Removed        []types.InspectionEnvironment
+	HashMismatches []HashMismatch
+}
+
+type HashMismatch struct {
```
matperin: nit: I think we can avoid exporting `DiffResult` and `HashMismatch`

microovn/api/inspect.go
matperin, line 1: Right now `inspect` is doing two jobs (handler logic and also all the parsing/probing logic).
The handlers should only need the top ~80 lines.
Do you think it would make sense to move the rest in a `microovn/api/inspect/` subpackage or, even better, in a `facts` package that both the API handlers and eventual future checks can reuse?
This would also make the `poll-until-converged` logic easier to find and test.

microovn/inspect/environment.go
```diff
@@ -120,6 +97,33 @@ func (EnvironmentCheck) Run(_ context.Context, input Input) []types.InspectionRe
 	}}
 }
 
+func nonAuthoritativeEnvironmentResult(input Input) types.InspectionResult {
+	result := types.InspectionResult{
```
matperin, line 95: nit: All results checks repeat the same prologue.
I would introduce a generic `forEachMemberSnapshot(input, factGroup, func(node, snapshot))` into `checks.go` (which, incidentally, would also make it more aptly named since right now it only holds two helpers and no checks) to consolidate the per-check boilerplate.

matperin (changes requested):
Thanks for the new feature @Deezzir!

I see that the "Network Routing and Datapath Validation" (scan the `DHCP` Options metadata parameters to ensure virtual machines maintain functional route declarations) seems to be missing on the diagnostic checks compared to what's written the specs. Was the drop of this step intentional?

Overall, good work on the PR, I left some comments on nits and minor structural change proposals.

I would also tidy up the commits structure a bit:
- [tests(unit): inspect command](https://github.com/canonical/microovn/pull/355/changes/740ef487ad5d3f8fed3cebfe7197445510ccf14d) can be squashed into [feat: inspect command](https://github.com/canonical/microovn/pull/355/changes/65f5adea835d3401342d23b7ac7fe72a83e8904d)
- [docs(inspect): add sudo for how-to docs](https://github.com/canonical/microovn/pull/355/changes/8314d53eebe78985c4f89e49ae80492fb0c3c851) can be squashed into [docs(inspect): document deployment inspection](https://github.com/canonical/microovn/pull/355/changes/6564768be70e756257a0fdf70cd22d2f04ec8d1a)
- [feat: finish topology check](https://github.com/canonical/microovn/pull/355/changes/9135f2a2846c7cc0b9b2267a114efe041a2b0a92), [feat: finish environment check](https://github.com/canonical/microovn/pull/355/changes/2dffa9448c4708834c4b3760d3e5727348dbbe33), [feat: finish environment check](https://github.com/canonical/microovn/pull/355/changes/2dffa9448c4708834c4b3760d3e5727348dbbe33) and [feat: finish database check](https://github.com/canonical/microovn/pull/355/changes/115f9d6045d10fa19e76c60d36c673695b13e560) can be squashed into [feat: inspect command](https://github.com/canonical/microovn/pull/355/changes/65f5adea835d3401342d23b7ac7fe72a83e8904d)
- likewise, [refactor: helpers for authoritative/desired check](https://github.com/canonical/microovn/pull/355/changes/0ce82355a3081d02e12f9b959fb1322c32d4fd17) and [refactor: split checks into separate files](https://github.com/canonical/microovn/pull/355/changes/b2a87c21fc32dd94307a09776d625542a2095efe) can also be squashed into [feat: inspect command](https://github.com/canonical/microovn/pull/355/changes/65f5adea835d3401342d23b7ac7fe72a83e8904d)
- The same is true for [fix: golangci-lint](https://github.com/canonical/microovn/pull/355/changes/88d5ed1ad25185736f1173476554eb3e96d5323b), [fix: correct error when no dqlite cluster](https://github.com/canonical/microovn/pull/355/changes/294b25ea7cb2d28c3a14edad636640af14f1cdc9), [fix: skip member specific en](https://github.com/canonical/microovn/pull/355/changes/3a3ce30e3ce39ba3a106867c63e5f81e8671f41d) and [fix: prettify the output](https://github.com/canonical/microovn/pull/355/changes/6cfff645e7048289331798201f1b9bb95fb59e9a)
- I would drop [fix: ignore sphinx build artifacts](https://github.com/canonical/microovn/pull/355/changes/55dd806019fb6f2cd458c62519a4cd58f1229f3e) from this PR and create a separate PR for this change since it is unrelated to the feature.
- Every remaining commit should also have a comment to summarize its contents. I know that the feature is pretty big but it would help referencing them in the future immensely. A summary is often enough, you do not need to get into the tiny details.

microovn/inspect/environment.go
Deezzir: Added https://github.com/Deezzir/microovn/blob/b0a2bd6b44fcca30219ccd295446d2bee27be178/microovn/inspect/checks.go#L16

microovn/api/inspect.go
Deezzir: Moved the parsing and probing to under `microovn/inspect/facts` - https://github.com/Deezzir/microovn/tree/feat/inspect/microovn/inspect/facts.

microovn/inspect/environment.go
Deezzir: Addressed

microovn/inspect/service_runtime.go
Deezzir: Addressed

https://github.com/Deezzir/microovn/blob/b0a2bd6b44fcca30219ccd295446d2bee27be178/microovn/inspect/service_runtime.go#L15-L27

microovn/inspect/collect.go
Deezzir: Addressed (folded in the switch below)

https://github.com/Deezzir/microovn/blob/b0a2bd6b44fcca30219ccd295446d2bee27be178/microovn/inspect/collect.go#L131-L149

microovn/api/inspect.go
Deezzir: Addressed:

https://github.com/Deezzir/microovn/blob/b0a2bd6b44fcca30219ccd295446d2bee27be178/microovn/inspect/facts/database.go#L69-L75
https://github.com/Deezzir/microovn/blob/b0a2bd6b44fcca30219ccd295446d2bee27be178/microovn/inspect/facts/database.go#L22-L30

Deezzir:
@matperin squashed into 3 commits:

[feat: inspect command](https://github.com/canonical/microovn/pull/355/commits/4a78f23ad9108e22dfe6fa2bb8fbfe101d50a3d6)
[docs(inspect): document deployment inspection](https://github.com/canonical/microovn/pull/355/commits/569d7da741e3cf1315411d18d2578ae1dea705f8)
[tests(func): cover inspect workflows](https://github.com/canonical/microovn/pull/355/commits/b0a2bd6b44fcca30219ccd295446d2bee27be178)

Also added summaries for each comment. I didn't see any recommendations for `tests(func): cover inspect workflows` so I left it as a third commit, please let me know if you want it squashed as well.

Deezzir:
@matperin Added Network/DHCP check:

- [`InspectionNetworkEndpoint`](https://github.com/Deezzir/microovn/blob/5ed8d69d0904795d2c40b18334710eb336d522bb/microovn/api/inspect.go#L34-L41)
- [`NetworkCheck`](https://github.com/Deezzir/microovn/blob/feat/inspect/microovn/inspect/network.go)

crypticC0der (approved):

matperin (approved):
Hi @Deezzir!
I apologize for the delay on my second review, thank you for the changes.
Everything seems in order now, I have no further comments.
LGTM!

